PR #24, opened 2015-01-24
Added german translation for input device settings.

ThomasVie:
Here is some more german translation.


jleclanche:
Landed in 536e296e9350a4079658b3f37565c7f2d4e58607

PR #25, opened 2015-01-24
Update lxqt-config-appearance_it.desktop

stefonarch:
Better using short names  because they  appear in LXQt global-systemsettings window and in menu


jleclanche:
Landed in 2916154

PR #26, opened 2015-01-24
Delete lxqt-config_it_IT.desktop

stefonarch:
double item


jleclanche:
Let's look at all these after 0.9, and clean them up together in all the lxqt modules.

PR #29, opened 2015-01-25
Update lxqt-config-input.ts

stefonarch:
Removed &amp; in "Add" and "Remove", because creates #_Add #_Remove in Gui

### DIFF
--- a/lxqt-config-input/translations/lxqt-config-input.ts
+++ b/lxqt-config-input/translations/lxqt-config-input.ts
@@ -86,12 +86,12 @@
     </message>
     <message>
         <location filename="../keyboardlayoutconfig.ui" line="63"/>
-        <source>&amp;Add</source>
+        <source>Add</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
         <location filename="../keyboardlayoutconfig.ui" line="70"/>
-        <source>&amp;Remove</source>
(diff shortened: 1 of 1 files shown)


jleclanche:
These are access keys, but I'm not sure why it would look like that on your setup. It looks fine on mine.

stefonarch:
I admit I ahve no idea what access keys means, but it looks liker that
(also in digikam I noticed the same thing).

Il 25/01/2015 15:51, Jerome Leclanche ha scritto:
> These are access keys, but I'm not sure why it would look like that on
> your setup. It looks fine on mine.
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/lxde/lxqt-config/pull/29#issuecomment-71375846>.
> 

-- 
********************************
Stef Binde
via della pieve 2
I-56024 Corazzano PI

Tel: ++39 0571 17 30 104
PGP: 0x728D9707

stef.binde@riseup.net
*******************************

jleclanche:
Access keys are when you press alt+key to select an item with the keyboard. They are an accessibility feature.

stefonarch:
Ok, got it.
So if I see them in source I let  them in the italian translation even
if it looks weird here?

Il 25/01/2015 18:02, Jerome Leclanche ha scritto:

PR #32, opened 2015-01-25
Update lxqt-config-cursor_it_IT.ts

stefonarch:
Hope no mistakes with filenames anymore...
Note: mouseover   to preview seems not working, has changed in this preview bar). And restart seems not necessary to take effects. Update translation source if thi is correct?


palinek:
landed in 718999e3036d978776c21776b8bdd8c2ee95bb8f

PR #39, opened 2015-02-03
- Fix desktop validation.

heliocastro:
Signed-off-by: Helio Chissini de Castro <helio@kde.org>


paulolieuthier:
@heliocastro thanks for your contributions! :smile:

PR #40, opened 2015-02-03
file-associations:Return from the busy cursor only after the Ui stuff ended

luis-pereira:
Processing the icons in a delayed batch mode is a great, great idea.
The cursor should be kept in the busy state until the Ui updating really
ends.


luis-pereira:
@PCMan A one liner change

paulolieuthier:
GTM.

luis-pereira:
Landed on lxde/lxqt-config@2a7ad099f160a7686ebc3522971a460ba21453b8

PR #48, opened 2015-06-02
Include X-LXQt category in lxqt-config menu

jubalh:
Some distributions patch our .desktop files to use' X-LXQt' instead of
'LXQt' to get Freedesktop conform desktop files.
See: http://sourceforge.net/p/lxde/mailman/message/34154860/

LXQt config only displays in the lxqt settings category if you have
category set to "LXQt" I ORed it to accept "X-LXQt" too.

The long term fix should be to get registered with Freedesktop.org

### DIFF
--- a/src/lxqt-config.menu
+++ b/src/lxqt-config.menu
@@ -19,6 +19,7 @@
                 <Category>DesktopSettings</Category>
                 <Or>
                     <Category>LXQt</Category>
+                    <Category>X-LXQt</Category>
                     <!-- Include some optional components here -->
                     <Filename>obconf-qt.desktop</Filename>
                     <Filename>compton-conf.desktop</Filename>


jleclanche:
At a glance I'd say that's correct but I want a 2nd opinion on this. ping @paulolieuthier @luis-pereira @lstranger

pmattern:
Wouldn't it be better to try and get 'LXQt' registered asap?
As jubalh already said 'X-LXQt' is actually only meant to get rid of again. So wouldn't introducing that string just mean superfluous work and a possible source of confusion?

jubalh:
I definitely think we should get registered, but I never dealt with freedesktop.org and have no idea how much of a hassle it is, and how long it takes. So I just created that change for now.
Although in openSUSE I am adding a patch that will do just the same (until there is a new release, or we got registered). To have it fixed quickly.
But I thought maybe other distributions have the same problem so I decided to provide this chagne for everyone.

Now it's your call :-)

jleclanche:
We're technically registered. The various XDG tools just suck. I sent an annoyed email the other day. There's not much else I can do.

jubalh:
Which tools? I could take a look at them if they accept patch from external people

pmattern:
@jubalh 
At least among other most of the scripts provided by xdg-utils, e. g. xdg-{desktop-icon,desktop-menu,email,open,mime}.

@jleclanche 
So there are two registrations, one to get into the xdg-foobar tools as stated above and another one to be recognized by keys OnlyShowIn and the like in the desktop entry files? And LXQt has already been registered for the first one but not yet for the latter?

LStranger:
In my opinion it is overall good solution, and in any case desired even
if and when LXQt is registered as official category. Although I see no
reason for XDG to add each DE into Categories list (currently only KDE,
GNOME, and XFCE are included, you know, no others DE) so doubt LXQt will
be there in nearest future.

LStranger:
And BTW, if someone missed the fact: the Categories value has nothing to
do with OnlyShowIn/NotShowIn value, those two are completely different
entities.

luis-pereira:
IMO that's the right way to do it. Technically,  it's correct http://standards.freedesktop.org/menu-spec/menu-spec-1.0.html#desktop-entry-extensions
I agree with @LStranger, I doubt that LXQt, will be added to the Categories.

> LXQt config only displays in the lxqt settings category if you have
category set to "LXQt" I ORed it to accept "X-LXQt" too.
@jubalh What are you referring to ?

paulolieuthier:
GTM.

pmattern:
FWIW https://bugs.freedesktop.org/show_bug.cgi?id=79561, https://bugs.freedesktop.org/show_bug.cgi?id=79562

jubalh:
@luis-pereira referring to what was the error and how i solved it

jubalh:
@pmattern very interesting. so the gentoo maintainer already asked them to do add LXQt too, even with a patch. I wonder why this task is still open...

luis-pereira:
@jubalh Sorry, but I'm really not understanding that sentence. Where is the OR ?

pmattern:
@luis-pereira 
I think he's referring to file lxqt-config.menu where both lines `<Category>LXQt</Category>` and `<Category>X-LXQt</Category>` would be in section `<Or>` after applying this PR.

luis-pereira:
@pmattern Thanks. Now I get it..... I only looked at the changed line.
GTM

jubalh:
*Establishing a string 'X-LXQt' means lots of superfluous work just because something as trivial as a registration of a rather well-known desktop environment hasn't been conducted since a year*

@pmattern 
I am not the one "establishing X-LXQt" it's used in many LXQt programs and configs already.

paulolieuthier:
@pmattern what do you suggest?

pmattern:
Personally I think another attempt should be made to evaluate whether it's possible to get the registration ahead.
If you consent I can look into this.
Depending on the results, which should be available in say one or two weeks, we could decide whether we merge this PR or we eventually won't need it any more.

paulolieuthier:
@pmattern feel free to try it, although I think it won't be that quick, as registration requests were already sent. We can hold this PR until the release time comes, then we merge it if there is no progress with the registration.

Putting more thought on it, I think even if we succeed registering, xdg's tools may take time to get to stable repositories of popular distributions, so it might be safer to merge this and eventually revert it.

LStranger:
@pmattern, you are messing up things I believe. Category has nothing in
common with Desktop Environment, read the standard, please. It is really
right to have Desktop Environment registered but that will NEVER change
the fact that there is no LXQt category, and as I already said I highly
doubt it will be ever added until you grow as big as KDE or GNOME. So in
short, add <Category>X-LXQt</Category> is pretty right thing to do, and
ever KDE has few X-* categories IIRC so don't be ashamed, please.

jubalh:
Merged with 7ee62db8fa8ce2f67f51d6040464f224580583df

PR #49, opened 2015-06-21
Updated german translation.

ThomasVie:
But... How to translate src/lxqt-config.menu:
<Name>LXQt settings</Name>
<Name>System settings</Name>
<Name>Other settings</Name>
resp.
<Menuname>LXQt settings</Menuname>
<Menuname>System settings</Menuname>

### DIFF
--- a/liblxqt-config-cursor/translations/lxqt-config-cursor_de.ts
+++ b/liblxqt-config-cursor/translations/lxqt-config-cursor_de.ts
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.0" language="de">
+<TS version="2.1" language="de">
 <context>
     <name>SelectWnd</name>
     <message>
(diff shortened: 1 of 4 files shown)


pmattern:
Regarding lxqt-config-monitor_de.ts I'd really urge you not to spend much time right now. It's under rather heavy development so it's likely that much has to be changed again anyway.

In lxqt-config_de.desktop I'd omit 'LXQt' in key Name[de] for the reasons stated in lxde/lxqt#713 but keep 'Konfigurationszentrum'. I mean neither this nor 'Konfigurations Center' are the real thing in German but I couldn't figure anything more suitable than the former.
I'm not happy at all with Comment[de] as 'modules' is used in a different meaning in lxqt-config-session - Basic settings. How about 'Konfiguration von LXQt- und Systemkomponenten'?

Aside from this GTM.

ThomasVie:
Made suggested changes in lxqt-config_de.desktop

palinek:
merged in @2cf4495, thanks

PR #54, opened 2015-07-29
Fix cursor settings location

City-busz:
Cursor settings are stored in session.conf, not in lxqt.conf.

Test case: try to set the cursor theme and relogin.


jleclanche:
gtm

PR #55, opened 2015-08-16
Initial support for locale config, based on formats KCMS (v2)

gilir:
This is an update of https://github.com/lxde/lxqt-config/pull/41

I rebased the changes on master,and added translation template.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ add_subdirectory(lxqt-config-input)
 add_subdirectory(lxqt-config-file-associations)
 add_subdirectory(lxqt-config-appearance)
 add_subdirectory(lxqt-config-monitor)
+add_subdirectory(lxqt-config-locale)
 
 # building tarball with CPack -------------------------------------------------
 include(InstallRequiredSystemLibraries)
--- a/lxqt-config-locale/CMakeLists.txt
+++ b/lxqt-config-locale/CMakeLists.txt
@@ -1,0 +1,58 @@
+project(lxqt-config-locale)
+
+set(H_FILES
+    localeconfig.h
+)
+
+
+set(CPP_FILES
+    main.cpp
+    localeconfig.cpp
+)
+
+set(UI_FILES
+    localeconfig.ui
+)
+
+# Translations **********************************
+lxqt_translate_ts(QM_FILES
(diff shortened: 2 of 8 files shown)


jleclanche:
I'll be testing it soon. @luis-pereira Can you test as well?

luis-pereira:
Sure @jleclanche

jleclanche:
Squash the first two commits, remove the man page, and this can go in imo.

gilir:
Thanks for the review. Update done.

jleclanche:
:+1: gtm!

paulolieuthier:
@gilir all occurrences of "LxQt" must be renamed to "LXQt" (or, rather, rebase to master). And IMO the region dropdown menu should be sorted alphabetically. There are also easily fixable compiler warnings. Other than that, GTM (although I haven't tested it very much).

gilir:
Thanks for the review. I fixed the LXQt naming and the compiler warning. Fixing the ordering seems more difficult, but as it's only a quick port of the KDE module, I think the UI can be improve after the initial support (having this very long combobox is probably not the best UI, but it's the case on the KDE module).

paulolieuthier:
Thanks.

pmattern:
Not sure to what degree the new configuration GUI is expected to work by now and this might be not the best place to report, but:

Running 5b50647 `lxqt-config-locale` doesn't have any effect on up to date Arch Linux x86_64 here. The locale that's chosen does get saved both in `~/.config/lxqt/LXQt-{config-locale,session}.conf`. Tested starting the session either from a virtual terminal via `startx` or from SDDM and using a locale that's different from the system's default locale and proven to work if invoked from a virtual terminal as `LANG=<locale> startx`.

Those list providing available locales seems to be hard-coded and has nothing to do with the locales that are in fact available in LXQt which can for sure be expected to cause confusion.
So I think in the long run the list should be adjusted accordingly.

luis-pereira:
@pmattern Those settings have to be exported to the environment. The best place to do it is the `startlxqt` script.

This is not the best place to do it, for sure, but.... why use LXQt-xxxxxxxxxx for config files ? lxqt-xxxxxxxxx is just fine IMO. We already have some discrepancies (some files have an lxqt prefix, others don't). Now we three naming "conventions".

PR #56, opened 2015-08-16
the license of qt-xcurtheme is WTFPL 2

agaida:
WTFPL 2 is a valid free non-permissive FOSS License acknowledged by the FSF


jleclanche:
gtm

agaida:
things are a little bit complicated - we are not done with only wtfpl 2

PR #57, opened 2015-08-18
License  headers and private library

agaida:
* make the included library private
* add missed license headers

### DIFF
--- a/liblxqt-config-cursor/CMakeLists.txt
+++ b/liblxqt-config-cursor/CMakeLists.txt
@@ -80,7 +80,7 @@ target_link_libraries(lxqt-config-cursor
 
 install(TARGETS
     lxqt-config-cursor
-    DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+    DESTINATION "${CMAKE_INSTALL_LIBDIR}/${PROJECT_NAME}"
     COMPONENT Runtime
 )
 install(FILES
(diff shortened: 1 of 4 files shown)


agaida:
@luis-pereira can you please have a look on the private .so patch?

i will rework the license headers for the 2 remaining files according to the mailing list.

agaida:
@luis-pereira, @paulolieuthier, @jleclanche ping

paulolieuthier:
AFAIK, the discussion in the ML has not settled.

agaida:
In my not so humble opinion - we can change this later, but i need some headers merged - without the headers i don't get the package into debian :)

paulolieuthier:
Then go ahead, I say it's GTM.

jleclanche:
Yeah this is fine for now.

liblxqt-config-cursor/CMakeLists.txt
```diff
@@ -80,7 +80,7 @@ target_link_libraries(lxqt-config-cursor
 
 install(TARGETS
     lxqt-config-cursor
-    DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+    DESTINATION "${CMAKE_INSTALL_LIBDIR}/${PROJECT_NAME}"
```
luis-pereira: @agaida Sorry for the delay. Real life. 
This causes an  error while loading shared libraries. The dynamic linker (default configuration) will not find the library there.
Filled an issue: https://github.com/lxde/lxqt/issues/793

PR #59, opened 2015-08-23
Updated and created italian .desktop files

stefonarch:


pmattern:
If I got this right you want to drop the country-specific files ("it_IT") from lxqt-config-appearance. As I don't speak Italian I don't know whether this is the right thing to do, IIRC we've discussed this before (don't remember where) so I'm just assuming it's okay.
But shouldn't this be consistent? There are lots of other components where country-specific files still exist, too, e. g. libxlqt-config-cursor, lxqt-config-monitor and lxqt-config (src/translations) in lxqt-config or at a glance lots of panel plugins.

stefonarch:
Yes we discussed that but I cannot find it now. I'm cleaning up  the files I stumble upon for now. There are only two   other  countries beside Italy, Switzerland (a small part in the south, Ticino) and the Rep. of San Marino, even smaller, so using only _it should fit all for now.
In this case _it.desktop was already translated well  and it_IT.desktop not.

pmattern:
> I'm cleaning up the files I stumble upon for now.

Alright.
If you don't mind I for one would find it good if you added the removal of / changes in those belonging to lxqt-config as stated above to this commit.

pmattern:
Merged now anyway (cd8f6f5) to get ahead.
Will remove the remaining country-specific stuff in another commit.

PR #62, opened 2015-10-30
Make the lxqt-config-locale chosen settings work

luis-pereira:
The settings were being written to the wrong config files.
Discussed at https://github.com/lxde/lxqt-config/pull/55.

### DIFF
--- a/lxqt-config-locale/main.cpp
+++ b/lxqt-config-locale/main.cpp
@@ -33,8 +33,8 @@
 int main (int argc, char **argv)
 {
     LXQt::SingleApplication app(argc, argv);
-    LXQt::Settings settings("LXQt-config-locale");
-    LXQt::Settings session_settings("LXQt-session");
+    LXQt::Settings settings("lxqt-config-locale");
+    LXQt::Settings session_settings("session");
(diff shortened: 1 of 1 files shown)


luis-pereira:
@pmattern Can you confirm that it's solves the problem you referred at https://github.com/lxde/lxqt-config/pull/55 ?

pmattern:
Running lxde/lxqt-config@562f25f plus all other LXQt components from this afternoon's (master) checkout `lxqt-config-locale` seems to work well now.

The LXQt sessions are using the locale set by `lxqt-config-locale`, no matter whether they're launched from the console by `startx` or from SDDM. Programs not belonging to LXQt like Otter Browser, QupZilla, Firefox or medit are respecting the locale, too.
Settings are now stored in `~/.config/lxqt/{session,lxqt-config-locale}.conf` only, so no capitalized names any longer.
What's still a problem IMO is the list of available languages shown by `lxqt-config-locale`. It has nothing to do with the languages that are in fact available which I'd suspect to give rise to a lot of trouble as stated in #55.

luis-pereira:
@pmattern I do agree with you about the available languages, Unfortunately  I will not have time to fix it before the release.
Shipping this.

PR #66, opened 2015-11-06
del:*_hu_HU.ts, upd:*_hu.ts, new:lxqt-config-locale_hu.ts

attuska:


pmattern:
In `lxqt-config-locale_hu.ts` strings "Imperial {UK,US}" and "Metric" aren't translated.

Other than that GTM.

Would be good if you could squash an eventual additional commit into the existing one (`git commit --amend` leaving the commit message as-is).

PR #73, opened 2015-12-05
A couple of bug fixes

f2404:
Check if the memory was allocated after malloc() and realloc() calls
Fix typo


paulolieuthier:
Merged in 6e9bfd8d2f2cd09c97b549b2552b6b33df130ac1. Thanks.

PR #76, opened 2016-01-23
lxqt-config: Load/store window size on begin/end

palinek:
closes lxde/lxqt#406

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -26,19 +26,29 @@
  * END_COMMON_COPYRIGHT_HEADER */
 
 #include <LXQt/SingleApplication>
+#include <QSettings>
 #include "mainwindow.h"
 
 
 int main(int argc, char **argv)
 {
     LXQt::SingleApplication app(argc, argv);
+    app.setOrganizationName(QStringLiteral("lxqt"));
+    app.setApplicationName(QStringLiteral("lxqt-config"));
 
     // ensure that we use lxqt-config.menu file.
     qputenv("XDG_MENU_PREFIX", "lxqt-");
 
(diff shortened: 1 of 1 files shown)


paulolieuthier:
GTM.

pmattern:
Window does remember its size once the latter got changed but upon start on a new user profile the window looks like this here
lxqt-config_window-size_first-launch

tsujan:
`QSettings::value()` (and perhaps, `QSettings::setValue()`) needs a reasonable fallback.

palinek:
@pmattern tried to fix in @790b577. Is it now OK?

pmattern:
As of 790b577 the size of the window of `lxqt-config` gets remembered, upon first start on a new user profile section "System settings" is still cut almost completely due to the window's small height.

Don't know whether the latter is something that really needs to get changed and tend to say no.
When parts of the window don't get displayed as its height is too low there's a scroll bar very well visible at the right edge. This plus the fact that changed window sizes do get remembered now should actually do, IMO.
Also, I wonder whether enforcing a larger height from the very beginning could eventually cause problems on rather tiny displays.

palinek:
> Don't know whether the latter is something that really needs to get changed and tend to say no.
When parts of the window don't get displayed as its height is too low there's a scroll bar very well visible at the right edge. This plus the fact that changed window sizes do get remembered now should actually do, IMO.
Also, I wonder whether enforcing a larger height from the very beginning could eventually cause problems on rather tiny displays.

Then, should we merge this? (I would say yes)

agaida:
I would suggest to ship it

palinek:
done

PR #78, opened 2016-03-01
Activate items on releasing Enter instead of pressing it.

tsujan:
Also take care of auto-repeating. This should fix https://github.com/lxde/lxqt/issues/965.


agaida:
GTM

PR #79, opened 2016-03-03
Set grid size according to font and icon sizes.

tsujan:
Fixes https://github.com/lxde/lxqt/issues/968.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -190,14 +190,15 @@ protected:
         QStyleOptionViewItemV4 opt = option;
         initStyleOption(&opt, index);
 
+        const QSize & iconSize = option.decorationSize;
         QSize size(mView->gridSize().width() - 8, // 4-px margin around each cell
-                   mView->iconSize().height());
+                   iconSize.height());
         // for having sharp non-scalable icons with HDPI
         int dpr = qApp->devicePixelRatio();
         if (dpr < 1) dpr = 1;
-        QPixmap pixmap = opt.icon.pixmap(mView->iconSize() / dpr); // -> Qt doc -> QIcon::pixmap()
-        if (dpr > 1 && pixmap.size() == mView->iconSize()) // exceptional (scalable or not from icon set)
-            pixmap = opt.icon.pixmap(mView->iconSize());
+        QPixmap pixmap = opt.icon.pixmap(iconSize / dpr); // -> Qt doc -> QIcon::pixmap()
+        if (dpr > 1 && pixmap.size() == iconSize) // exceptional (scalable or not from icon set)
+            pixmap = opt.icon.pixmap(iconSize);
         opt.icon = QIcon(pixmap.copy(QRect(QPoint(0, 0), size * dpr)));
         opt.decorationSize = size;
 
@@ -220,7 +221,6 @@ LXQtConfig::MainWindow::MainWindow() : QMainWindow()
     model = new ConfigPaneModel();
 
     view->setViewMode(QListView::IconMode);
-    setSizing();
     view->setWordWrap(true);
     view->setUniformItemSizes(true);
     view->setCategoryDrawer(new QCategoryDrawerV3(view));
@@ -228,6 +228,7 @@ LXQtConfig::MainWindow::MainWindow() : QMainWindow()
     connect(view, &QAbstractItemView::activated, this, &MainWindow::activateItem);
     view->setFocus();
 
+    QTimer::singleShot(0, [this] { setSizing(); });
     QTimer::singleShot(1, this, SLOT(load()));
     new QShortcut{QKeySequence{Qt::CTRL + Qt::Key_Q}, this, SLOT(close())};
 }
@@ -260,8 +261,7 @@ void LXQtConfig::MainWindow::activateItem(const QModelIndex &index)
 void LXQtConfig::MainWindow::setSizing()
 {
     // consult the style to know the icon size
-    int iconSize = qBound(16, QApplication::style()->pixelMetric(QStyle::PM_IconViewIconSize), 256);
-    view->setIconSize(QSize(iconSize, iconSize));
+    int iconSize = qBound(16, view->decorationSize().height(), 256);
     // DPR is automatically taken into account in setIconSize()
     // but wee need to consider it explicitly below
     int dpr = qApp->devicePixelRatio();
--- a/src/qcategorizedview/qcategorizedview.cpp
+++ b/src/qcategorizedview/qcategorizedview.cpp
@@ -781,6 +781,11 @@ void QCategorizedView::reset()
     QListView::reset();
 }
 
+QSize QCategorizedView::decorationSize() const
+{
+    return viewOptions().decorationSize;
+}
+
 void QCategorizedView::paintEvent(QPaintEvent *event)
 {
     if (!d->isCategorized()) {
--- a/src/qcategorizedview/qcategorizedview.h
+++ b/src/qcategorizedview/qcategorizedview.h
@@ -213,6 +213,11 @@ public:
       */
     virtual void reset();
 
+    /**
+     * @return The icon size by considering all the styling
+     */
+    QSize decorationSize() const;
+
 protected:
     /**
       * Reimplemented from QWidget.


palinek:
Although not optimal (not considering the real texts)...this is a good approach.

Just one note... the sizing isn't changed on the fly with changes in lxqt-config-appearance.

@tsujan consider the commit @adb806a

GTM

tsujan:
> the sizing isn't changed on the fly...

I'll look into it.

> ... consider the commit...

OK, I'll give it a separate method.

> not considering the real texts

The real text could be considered in the same way it is in libfm-qt but I think `QCommonStyle` draws it more elegantly with `drawControl()`. Moreover, it's better to leave drawings to the style engine as far as possible. (I have a plan to replace libfm-qt's method with that of `QCommonStyle` too.)

tsujan:
There was a minor issue, that I fixed with the third commit:

At least with some fonts or font sizes, when the window started, the position of item texts was at top but if the window was maximized and then unmaximized again, they were positioned vertically centered.

The reason was that the listview was shown before it was fully formed. In fact, *the default positioning of text was always vertically centered* and, I guess, you'd tried to enforce the top position by using `int height` inside `sizeHint()` (which I removed). If you still prefer the top position, this could be added to `paint()`:
````
opt.displayAlignment = Qt::AlignHCenter | Qt::AlignTop;
````
However, I found the centered position more elegant.

**Comment by tsujan** [comment]
Seeing that KDE systemsettings aligns texts to top, I came up with a better calculation of `sizeHint()`. Hopefully, all is OK now.

**Comment by palinek** [comment]
> I guess, you'd tried to enforce the top position by using int height inside sizeHint() (which I removed).

Based on git-blame.... this is the commit with reasoning @20b5867 -> Razor-qt/razor-qt#530

**Comment by tsujan** [comment]
> Based on git-blame....

That could hardly happen again. In https://github.com/lxde/lxqt-config/commit/d9ce1c9d57e14329a6dc2fab0bf4ea306a16a34f, I came to almost the same method (after seeing systemsettings) because it had an advantage.

**Comment by palinek** [comment]
@tsujan
How about removing the whole homemade `ConfigItemDelegate` at all and letting the qt handle it? Isn't defining the size grid enough?

The item delegate is broken. Compare
- QT_DEVICE_PIXEL_RATIO=1

screenshot
- QT_DEVICE_PIXEL_RATIO=2 (also with duplicated font DPI: `echo 'Xft.dpi: 192' | xrdb -merge -`)

screenshot

- QT_DEVICE_PIXEL_RATIO=2 (also with duplicated font DPI: `echo 'Xft.dpi: 192' | xrdb -merge -`, but without usage of `ConfigItemDelegate`)
screenshot

**Comment by tsujan** [comment]
> How about removing the whole homemade ConfigItemDelegate at all and letting the qt handle it?

I agree about letting Qt (and especially, QStyle) do such jobs as far as possible. If you think the removal of `ConfigItemDelegate` will have no side effect, do it in another PR and we could test it. If successful, we'll get rid of `ConfigItemDelegate`; if not, I'll fix the HDPI problem here :)

**Comment by tsujan** [comment]
HDPI is fixed now. This is with `export QT_DEVICE_PIXEL_RATIO=2`, a png icon set and gtk+ style:

dpi

No blurred icon.

**Comment by tsujan** [comment]
All is OK now. With 192-dpi font:
with_192_dpi

**Comment by palinek** [comment]
This is what I see now with QT_DEVICE_PIXEL_RATIO=2
- 11pt font (96DPI):
screenshot

- 11pt font (192 DPI):
screenshot

**Comment by tsujan** [comment]
> This is what I see now with QT_DEVICE_PIXEL_RATIO=2

A link for testing your icon theme?

**Comment by tsujan** [comment]
I asked because this is with Gnome icon theme:
2016-03-15-16 33 02-scrot

**Comment by tsujan** [comment]
BTW and not related to this discussion, icons in `/usr/share/pixmaps` are still broken like the day one I came to LXQT. It's a Qt5 bug, I guess.

**Comment by palinek** [comment]
> BTW and not related to this discussion, icons in /usr/share/pixmaps are still broken like the day one I came to LXQT. It's a Qt5 bug, I guess.

It's been fixed recently -> lxde/libqtxdg@18a4abf

**Comment by palinek** [comment]
> A link for testing your icon theme?

I'm using the numix circle light.

GNOME icon theme QT_DEVICE_PIXEL_RATIO=2
- 11pt font (96DPI):

screenshot
- 11pt font (192 DPI):

screenshot

**Comment by tsujan** [comment]
The brightness icon is still SVG, I suppose.

Anyhow, I had an SVG set and saw the same thing. Once I removed the line `app.setAttribute(Qt::AA_UseHighDpiPixmaps, true);` from `main.cpp` (without changing anything else in this PR), they were OK.

Now, this is the problem to solve with HDPI: On the one hand, with this PR and with SVG icons, the icons are cut; on the other hand, with https://github.com/lxde/lxqt-config/pull/81 and with PNG icons, they're blurred.

I try to find a solution here or at https://github.com/lxde/lxqt-config/pull/81, whichever is easier.

**Comment by tsujan** [comment]
The last problem is also solved by a trick I used in Kvantum for similar situations. Screenshot with Numix:
numix
Since we're dealing with a multi-headed dragon, please test!

**Comment by palinek** [comment]
Yes, it seems to be fixed... but do you still think it is worth the pain? As now the behaviour is the very same as without the `ConfigItemDelegate`, not?

**Comment by tsujan** [comment]
> Yes, it seems to be fixed

Ah, thanks! Now I understand why most devs hate QStyle ;)

> As now the behaviour is the very same as without the ConfigItemDelegate, not?

There are two differences, of which the first one is important:

(1) PNG icons would be blurred in the item view without `ConfigItemDelegate`. The whole point of HDPI is to have sharp big icons and text.

(2) With `ConfigItemDelegate`, The HDPI margins and spacings are respected (they're double-sized, of course). If it were only about this, we could get rid of `ConfigItemDelegate` because it wouldn't be a big deal.

**Comment by palinek** [comment]
> (1) PNG icons would be blurred in the item view without ConfigItemDelegate. The whole point of HDPI is to have sharp big icons and text.

Well... then this is a bug in Qt and should be reported... I don't think we can/should fix it in every possible widget/view that we are using in lxqt.

**Comment by tsujan** [comment]
> Well... then this is a bug in Qt

I don't think so because I can't see such issues in "pure" listviews or treeviews. . When we subclass an item view, we should take care of size calculations, of which `QStyledItemDelegate` is a part. Otherwise, unexpected things might happen, as they did in the case of PCManFM (and fixed months ago).

**Comment by tsujan** [comment]
@palinek with https://github.com/lxde/lxqt-config/pull/79/commits/13e91e75d3f2af2dc5666b4ea9647b3f43cb816f, there's no icon at all:
2016-04-15-15 35 09-scrot

**Comment by tsujan** [comment]
Did you have a specific reason to remove `view->setIconSize(QSize(iconSize, iconSize));`? Without it, icons wouldn't be displayed.

**Comment by palinek** [comment]
> Did you have a specific reason to remove view->setIconSize(QSize(iconSize, iconSize));?

Yes, it was intentional.

>  Without it, icons wouldn't be displayed.

Why? If you don't override the icon size, the style's value will be used.

Don't know why you are missing the icons, here they are shown correctly:
screenshot

Could Kvantum somehow interfere? (https://github.com/tsujan/Kvantum/commit/75ab6acf4f74ba745b7aa9611997c5c3ab8c83f6)

**Comment by tsujan** [comment]
> Could Kvantum somehow interfere?

I get no icon with fusion either:
2016-04-15-16 21 07-scrot

> Yes, it was intentional.

But why? With it, `lxqt-config.qss` works OK.

**Comment by palinek** [comment]
Is there any difference if you remove the specific `itemDelegate`?:
```diff
diff --git a/src/mainwindow.cpp b/src/mainwindow.cpp
index ebdedad..2b49986 100644
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -242,7 +242,7 @@ void LXQtConfig::MainWindow::load()
     proxyModel->setSourceModel(model);
 
     view->setModel(proxyModel);
-    view->setItemDelegate(new ConfigItemDelegate(view));
+    //view->setItemDelegate(new ConfigItemDelegate(view));
 
     view->show();
 
```

tsujan:
> Why? If you don't override the icon size, the style's value will be used.

Because the view is subclassed. I still don't understand how you have those icons!!

tsujan:
> Is there any difference if you remove the specific itemDelegate?

Of course, it does. But why removing it and not re-adding `view->setIconSize` ? If there's no `qss` file or the icon size isn't set in it, the style's value will be used automatically. Try it!

palinek:
> > Why? If you don't override the icon size, the style's value will be used.

> Because the view is subclassed.

I don't follow... why should it make a difference?

tsujan:
You're worried about the style's value not being used? As I said above, it will be used. Please just try that!

palinek:
> If there's no qss file or the icon size isn't set in it, the style's value will be used automatically. Try it!

I know... 

> ... not re-adding view->setIconSize ?

Because it shouldn't be needed (the `iconSize` property is used to override the style, but we aren't overriding it... we want to use the style's value). If you don't see the icons w/o the `setIconSize` there is a problem somewhere in the code.

palinek:
> You're worried about the style's value not being used?

Nope.  It is like saying "override = default" .... what is the purpose of such override?

tsujan:
Sorry, you were right! I'd changed the code manually, based on the general concept, and missed a line. Once I applied the patch, icons were displayed.

> It is like saying "override = default" .... what is the purpose of such override?

That's true. However, I remembered examples, where QCommonStyle left a job to the overriding class, hence I got too cautious.

In short, yes, `setIconSize` can be removed, as you did. Everything looks OK.

tsujan:
BTW, Qt's HDPI has some problems with stylesheets as well as with odd DPIs ≥ 3. They're Qt bugs and aren't our concern.

palinek:
As no one except the two of us takes care of this...  I'm +1 on merging this. How about you?

tsujan:
> How about you?

Me too. Let's get some feedback! If there's a problem, it shouldn't be serious and could be fixed by a separate, clean PR.

PR #81, opened 2016-03-14
lxqt-config: Remove specific ConfigItemDelegate

palinek:
This removal fixes problem depicted in https://github.com/lxde/lxqt-config/pull/79#issuecomment-196280378

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -167,11 +167,14 @@ LXQtConfig::MainWindow::MainWindow() : QMainWindow()
 {
     setupUi(this);
 
+    /* To always have the intended layout with a vertically centered text
+       on startup, the listview should be shown after it's fully formed. */
+    view->hide();
+
     model = new ConfigPaneModel();
 
     view->setViewMode(QListView::IconMode);
-    view->setIconSize(QSize(32, 32));
-    view->setGridSize(QSize(100, 100));
+    setSizing();
     view->setWordWrap(true);
     view->setUniformItemSizes(true);
     view->setCategoryDrawer(new QCategoryDrawerV3(view));
@@ -193,6 +196,8 @@ void LXQtConfig::MainWindow::load()
 
     view->setModel(proxyModel);
 
+    view->show();
+
     QApplication::restoreOverrideCursor();
 }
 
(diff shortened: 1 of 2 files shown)


tsujan:
I can't see any problem, especially when https://github.com/lxde/lxqt-config/pull/79 is also used. Would you please mix https://github.com/lxde/lxqt-config/pull/79 and this PR together to have the best of both in the same PR?

Just a question: Why `ConfigItemDelegate` was introduced, in the first place?

palinek:
> Just a question: Why ConfigItemDelegate was introduced, in the first place?

I've mentioned it here: https://github.com/lxde/lxqt-config/pull/79issuecomment-192125566

tsujan:
There's just one thing: selection margins aren't respected for styles that, unlike Fusion, highlight the whole item (gtk+ is an example). In fact, as far as I see, that was the only use of `ConfigItemDelegate`. It isn't a big deal but if we want to add such a margin later, we'll have to reintroduce `ConfigItemDelegate`.

tsujan:
The link you gave is broken.

tsujan:
As an example I what I said, see these screenshots with the gtk+ style:

2016-03-14-21 26 00-scrot
2016-03-14-21 26 15-scrot

The selection margins are respected in the second screenshot, which is taken without this PR but with #79.

palinek:
> I can't see any problem, especially when #79 is also used. Would you please mix #79 and this PR together to have the best of both in the same PR?

Merged here.



> The link you gave is broken.

Fixed right after submit.

tsujan:
OK! Now I understand why it was added. I don't think that's an issue anymore.

What do you think about https://github.com/lxde/lxqt-config/pull/81#issuecomment-196442547?

palinek:
> What do you think about #81 (comment)?

I would vote for simplicity/removal, but i think you are more experienced in this area to judge it. I've never dealt with this kind of styling/hinting etc. So feel free to make the decision on your own.

tsujan:
@palinek 
I guess the icon set you used in https://github.com/lxde/lxqt-config/pull/79#issuecomment-196280378 was SVG. I repeated the test with a PNG icon set (and without https://github.com/lxde/lxqt-config/commit/7a6312d1c3f91634831a0d220e97d5e5f41d6bff, as you did before) and saw that big icons were blurred, *even when I added `app.setAttribute(Qt::AA_UseHighDpiPixmaps, true);` to `main.cpp`* . Do you confirm this?

And if you consider https://github.com/lxde/lxqt-config/commit/7a6312d1c3f91634831a0d220e97d5e5f41d6bff, icons won't be enlarged at all because their size will be set in setSizing().

It seems that we can't get rid of `ConfigItemDelegate` and have a consistent appearance. IMO, That's because we have to use `QCategorizedView`.

I have a simple way out of this, which I'll add to https://github.com/lxde/lxqt-config/pull/79.

tsujan:
I was wrong when I said: "And if you consider 7a6312d, icons won't be enlarged at all". The correct icon size is set with `echo 'Xft.dpi: 192' | xrdb -merge`. However, the icons are still blurry.

I'm not satisfied with https://github.com/lxde/lxqt-config/commit/2439ceac4776ee3a894b9420acf37f3ef865965e either -- although it fixes the blurring issue, it's just a workaround.

Let's think more about these problems.

tsujan:
OK, now #79 is correct both theoretically and practically: the sizes, spacings and margins are correct for both QT_DEVICE_PIXEL_RATIO=1 and QT_DEVICE_PIXEL_RATIO=2, while the icons remain sharp in the latter case.

What had confused me was why Fusion's view icon size depended on font DPI but the answer was in some Qt private headers.

PR #83, opened 2016-04-15
font config: make it possible to write custom fontconfig files

RalfJung:
Right now, lxqt-config-appearance overwrites the `fonts.conf` file. That is rather annoying if you have other options in that file that lxqt-config does not know or understand (like fine-tuning the font aliasing).

This commit adds an "include" line to the generated `fonts.conf` so that one can drop custom files in `~/.config/fontconfig/conf.d`. According to `man fonts.conf`, that is a standard location, but it still has to be explicitly included by the `fonts.conf`.


jleclanche:
gtm

palinek:
GTM

PR #86, opened 2016-04-24
Translation QMs were not loaded for brightness settings and few notr …

Ilya87:
…for Form


palinek:
Gtm

paulolieuthier:
GTM.

palinek:
landed in @1eb976a, thanks

PR #90, opened 2016-05-18
lxqt-config-appearance: Use the new XdgIconLoader library

luis-pereira:
Just adapting to it. Nothing new.
To be merged after lxde/libqtxdg#84 gets in.


agaida:
GTM

paulolieuthier:
GTM.

palinek:
Actualy, here IMO we shouldn't use the Xdg stuff for the theme change.... just fiddling with that... will submit PRs soon.

PR #91, opened 2016-05-19
fixed crash when selecting "extended view" while second monitor is off

qsodev:


lxqt-config-monitor/fastmenu.cpp
```diff
@@ -58,7 +58,16 @@ void FastMenu::extended()
         pos.setY(0);
         output->setPos(pos);
         output->setEnabled(true);
-        width += output->currentMode()->size().width();
+        //first left one as primary
+        output->setPrimary(width == 0);
+        KScreen::ModePtr mode(output->currentMode());
+        if (!mode) { // set first mode
```
palinek: Please, use styling (open/close brace on new line) as it is used in the surrounding code.

palinek:
GTM (with the styling change)

qsodev:
GTM or are other changes needed?

PR #92, opened 2016-05-20
lxqt-config-appearance: Use fewer XdgIcon* stuff

palinek:
As the XdgIconLoader is now used as backend for all QIcon stuff (if lxqt-qtplugin is used), use the XdgIcon* stuf as least as possible -> let the lxqt-qtplugin & Qt handle it.


luis-pereira:
GTM

paulolieuthier:
GTM.

agaida:
GTM

PR #93, opened 2016-06-01
lxqt-config-appearance: Fix a null pointer dereference

luis-pereira:
If (distance < minimalSize) is never true, we end up dereferencing
closestMatch that was initialized to a null pointer.


paulolieuthier:
GTM.

agaida:
GTM

PR #94, opened 2016-07-10
I18n: Fix Plural String

SafaAlfulaij:
The "time remaining" string isn't treated as plural.
I hope I did it right! :smile:


lxqt-config-monitor/timeoutdialog.cpp
```diff
@@ -59,7 +59,7 @@ void TimeoutDialog::onTimeout()
     else
     {
         int remaining = maximum / TIMER_DURATION - TIMER_DURATION * time / maximum;
-        ui.remainingTime->setText(tr("%1 second(s) remaining").arg(remaining));
+        ui.remainingTime->setText(tr("%n second(s) remaining", NULL, remaining).arg(remaining));
```
paulolieuthier: Please, use `nullptr` instead of `NULL`.

PR #95, opened 2016-07-18
lxqt-config-monitor: Use QGuiApplication

palinek:
From the libkscreen v5.7(?), the KScreen::GetConfigOperation() does not
provide any outputs information if only the QCoreApplication is used.

Fixes lxde/lxqt#1096


pmattern:
Problem is fixed in ab9543f.
Both single and dual monitor systems, both hardware and QEMU virtual machines.

PR #99, opened 2016-09-06
appearance: Do (not) change wallpaper with theme

palinek:
When chaning the LXQt theme, the wallpaper will be changed only if it
wasn't changed by the user independently (the config value
wallpaperChanged is set by pcmanfm-qt). But we can override chaning the
wallpaper (new checkbox added).

This needs lxde/pcmanfm-qt#396

closes lxde/lxqt#156


paulolieuthier:
GTM.

jubalh:
I think this definitely should be in the new release.
I would like to add the label so we dont forget to wait for this merge. Ok?

palinek:
I've changed the milestone in the issue....

palinek:
...rebased&squashed.

Are we going to merge these changes even if it needs the pcmanfm-qt help for the configs?

paulolieuthier:
No, let's wait for the other PR.

PR #100, opened 2016-09-13
brightness: Add change confirmation dialog

palinek:
After settings change (1 sec restartable timeout) the confirmation is
requested (confirmation checkbox is default checked). The dialog will
timeout/cancel automaticaly (after 5 seconds) if the user doesn't
confirm it.

By this we are avoiding unrecoverable setting (full white/black screen).

closes lxde/lxqt#973

### DIFF
--- a/lxqt-config-brightness/brightnesssettings.cpp
+++ b/lxqt-config-brightness/brightnesssettings.cpp
@@ -18,7 +18,8 @@
 
 #include "brightnesssettings.h"
 #include "outputwidget.h"
-#include <QDebug>
+#include <QMessageBox>
+#include <QPushButton>
 
 BrightnessSettings::BrightnessSettings(QWidget *parent):QDialog(parent)
 {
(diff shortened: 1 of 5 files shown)


stefonarch:
Works as espected here. See screenshots:
schermata
schermata01

lxqt-config-brightness/brightnesssettings.cpp
```diff
@@ -18,30 +18,79 @@
...
+                m.setBrightness(monitor.brightness());
+            }
+        }
+    }
 }
 
+void BrightnessSettings::requestConfirmation()
+{
+    QMessageBox msg{QMessageBox::Question, tr("Brightness settings changed")
+        , tr("Confirmatis required. Are the settings correct?")
```
paulolieuthier: Typo and incorrect alignment.

paulolieuthier:
GTM.

palinek:
Fixed typo & slightly updated the logic (using the "No" text from Qt QMessageBox::StandardButton)

PR #101, opened 2016-09-13
Add Qt5XdgIconLoader as dependency

jubalh:
Used in lxqt-config-appearance/iconthemeinfo.cpp.

PR #102, opened 2016-09-14
all: Remove XdgIcon & XdgIconLoader usage

palinek:
We don't need to use the XdgIcon any more, as the XdgIconLoader is used
as the default QIconEngine (thanks to our platform theme plugin -
lxqt-qtplugin).

For the very same reason, the (duplicated) icon searching logic was
removed from lxqt-config-appearance/iconthemeinfor.cpp.

closes #101


paulolieuthier:
GTM.

PR #103, opened 2016-09-14
Wallpaper theme2

palinek:
When chaning the LXQt theme, the wallpaper will be changed only if it
wasn't changed by the user independently (checking 'Desktop/Wallpaper'
key in pcmanfm-qt's config). But we can override changing the wallpaper
(new checkbox added).

closes lxde/lxqt#156
closes lxde/lxqt-config#99
closes lxde/pcmanfm-qt#396

This is another implementation w/o the need of pcmanfm-qt's assistance, as suggested in https://github.com/lxde/pcmanfm-qt/pull/396#issuecomment-247134654

### DIFF
--- a/lxqt-config-appearance/lxqtthemeconfig.ui
+++ b/lxqt-config-appearance/lxqtthemeconfig.ui
@@ -51,6 +51,13 @@
      </column>
     </widget>
    </item>
+   <item>
+    <widget class="QCheckBox" name="wallpaperOverride">
+     <property name="text">
+      <string>Override user-defined wallpaper</string>
+     </property>
+    </widget>
+   </item>
   </layout>
  </widget>
  <resources/>
--- a/lxqt-config-appearance/lxqtthemeconfig.cpp
+++ b/lxqt-config-appearance/lxqtthemeconfig.cpp
@@ -144,11 +144,9 @@ void LXQtThemeConfig::lxqtThemeSelected(QTreeWidgetItem* item, int column)
 		QString wallpaper = theme.desktopBackground();
 		if(!wallpaper.isEmpty() && (ui->wallpaperOverride->isChecked() || !isWallpaperChanged(currentTheme.desktopBackground()))) {
 			// call pcmanfm-qt to update wallpaper
-			QProcess process;
 			QStringList args;
 			args << "--set-wallpaper" << wallpaper;
-			process.start("pcmanfm-qt", args, QIODevice::NotOpen);
-			process.waitForFinished();
+			QProcess::startDetached("pcmanfm-qt", args);
 		}
 	}
 }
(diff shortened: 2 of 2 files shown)


lxqt-config-appearance/lxqtthemeconfig.ui
```diff
@@ -51,6 +51,13 @@
      </column>
     </widget>
    </item>
+   <item>
+    <widget class="QCheckBox" name="wallpaperOverride">
+     <property name="text">
+      <string>Override independently set wallper</string>
```
paulolieuthier: Typo.

tsujan:
Simple and effective! After fixing the typo, GTM.

lxqt-config-appearance/lxqtthemeconfig.ui
tsujan:
````
<string>Always use theme wallpaper</string>
````

palinek: How about 'Force theme wallpaper' ?

paulolieuthier: "Use theme's wallpaper when changing themes.", or something like that. "Forcing" would seem like the user could't change it.

tsujan: I agree with @paulolieuthier. Variations:

Use theme's wallpaper
Use theme wallpaper
Wallpaper from the theme

palinek: But it's not only about using it... it's about overriding... 
Even when unchecked the wallpaper is changed (if the pre-existing wallpaper is set by the theme).

paulolieuthier: @palinek you're right. Maybe "Override user-defined wallpaper when selecting themes".

palinek: Isn't the "Override user-defined wallpaper" enough?

tsujan: IMO, it is. Change it and merge the PR!

paulolieuthier: Yes, it is. 👌

jeremywh7: I think that might be confusing to some users; what about something like:
> Allow theme to use default wallpaper

Sorry to add more fuel to this fire...  :-P

tsujan: @jeremywh7  Most of these phrases are good. Your chosen phrase would be better in this form: "Allow theme to *set* the wallpaper". However @palinek intends to emphasise overriding.

PR #111, opened 2016-12-20
Fix some warnings

luis-pereira:


liblxqt-config-cursor/xcr/xcrimg.cpp
```diff
@@ -59,9 +59,9 @@ void XCursorImage::convertARGB2PreMul (QImage &img) {
   switch (img.format()) {
     case QImage::Format_ARGB32_Premultiplied: return;
     case QImage::Format_ARGB32: break;
-    default: img.convertToFormat(QImage::Format_ARGB32/*_Premultiplied*/);
+    default: (void) img.convertToFormat(QImage::Format_ARGB32/*_Premultiplied*/);
```
palinek: static_cast<void>(...) ?

palinek: Sure, I'm just nitpicking about the "clean" C++ style... as the `(blah)` casting notation is the C style... 

Till now I was in impression, that the `-Wold-style-cast` will trigger warning also with the `(void)` on return value, but it is not the case for either clang v3.8 or g++ v6.2. So we can forget my comment.

luis-pereira: @palinek I understood. With g++ v6.2 and clang v3.9 we don't get the -Wold-style-cast.
`(void)(...)` is the only case that doesn't trigger the warning.

The standard says:
> Any expression can be explicitly converted to type “cv void.” The expression value is discarded.

The cast isn't really a cast. Is more of a discard.

PR #112, opened 2016-12-26
make use of lxqt-build-tools

tpgxyz:
This PR adds support for lxqt-build-tools.
Basically this is needed to have consistent path for LXQT_ETC_XDG_DIR

For example:

LXQT_ETC_XDG_DIR set by lxqt-build-tools is /etc/xdg/qt5
LXQT_ETC_XDG_DIR set by lxqt-config is set to /etc/xdg

running lxqt-config gives error as it cannot find lxqt-config.menu

@luis-pereira Please review and merge

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.0.2 FATAL_ERROR)
 
 project(lxqt-config)
 set(LXQT_CONFIG_PROJECT "${PROJECT_NAME}")
+set(LXQTBT_MINIMUM_VERSION "0.1.0")
 
 option(UPDATE_TRANSLATIONS "Update source translation translations/*.ts files" OFF)
 option(WITH_INPUT "Build the 'lxqt-config-input'" ON)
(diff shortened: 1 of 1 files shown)


luis-pereira:
@tpgxyz 
`lxqt-build-tools` is a` liblxqt` dependency. And it already includes `LXQtConfigVars'.
Check it [here](https://github.com/lxde/liblxqt/blob/master/cmake/lxqt-config.cmake.in#L17).

tpgxyz:
@luis-pereira This commit is not available in liblxqt-0.11.0 release.
I'd suggest to do a minor release of liblxqt when people start whining :)

luis-pereira:
That's something for @agaida to consider.

agaida:
i have no problem to push the big red button right now :) - but wait, it would be nice if someone could review and test lxde/lxqt-panel/pull/387 before - i would really like to have it in.

@tpgxyz - your whish is absolutely understandable and reasonable. And indeed the build tools are planned right after the 0.11.0 release to make partial releases and updates possible.

luis-pereira:
@agaida Only tomorrow I have a chance to take a look at https://github.com/lxde/lxqt-panel/pull/387

agaida:
@luis-pereira - no hurry - i thought in the mean time about it again. Maybe it is better to roll out the current state and switch this PR to a later time. (Release it with the other things like DnD) - This would be around Jan. 15

tpgxyz:
Thanks

PR #114, opened 2016-12-31
lxqt-config-monitor: remove unnecessary wait before settings load

10ne1:
Recent versions of kscreen (tested with latest 5.8.5 and also with 5.6)
don't have any noticeable slowness so remove this random magic number wait
because it's very anoying when using lxqt-config-monitor -l to restore
previously saved configurations.

Signed-off-by: Ioan-Adrian Ratiu <adi@adirat.com>


palinek:
Full ack... and to be honest, the lxqt-config-monitor needs more cleaning...

agaida:
GTM

PR #115, opened 2017-01-17
Use const iterators

luis-pereira:
Avoid assignment of a non-const pointer to a const pointer and the
consequent container detach.


jleclanche:
gtm

PR #116, opened 2017-02-23
Show non-primary screens in UI.

selairi:
Fix a simple bug in UI. Extend display option is shown when screen is not a primary screen:

captura


pmattern:
Did I get it right that the purpose of this PR is to *automatically* select "This screen extends another display", so if one display is deliberately declared the primary, then all others are automatically assigned "This screen extends..."?

If so, it works as expected here and doesn't seem to introduce any problems. So GTM.

selairi:
Did I get it right that the purpose of this PR is to *automatically* select
> "This screen extends another display", so if one display is deliberately
> declared the primary, then all others are automatically assigned "This
> screen extends..."?
>
>
Yes, this is the idea.

agaida:
GTM

PR #117, opened 2017-04-18
Fix build with CMake 3.8

yan12125:
In CMake 3.8, the path to generated ui_*.h files is changed.

Fixes https://github.com/lxde/lxqt/issues/1277

The new path is documented, [1] so I hard-coded it. The current approach looks ugly as there are duplicated codes. Unfortunately I'm not familiar with CMake enough to give an elegant fix.

```
The generated generated ui_*.h files are placed in the <CMAKE_CURRENT_BINARY_DIR>/<TARGETNAME>_autogen/include directory which is automatically added to the target’s INCLUDE_DIRECTORIES.
```
[1] https://cmake.org/cmake/help/v3.8/manual/cmake-qt.7.html#autouic

### DIFF
--- a/liblxqt-config-cursor/CMakeLists.txt
+++ b/liblxqt-config-cursor/CMakeLists.txt
@@ -74,6 +74,16 @@ add_library(lxqt-config-cursor
         ${lxqt-config-cursor_QM_LOADER}
 )
 
+if(CMAKE_VERSION VERSION_LESS 3.8.0)
+    set(UI_PATH "${CMAKE_CURRENT_BINARY_DIR}")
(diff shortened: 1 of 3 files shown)


palinek:
GTM

luis-pereira:
@yan12125 My proposal: https://github.com/lxde/lxqt-config/pull/118

The CMake is just a symptom, not the root cause. 
But the cmake stuff needs maintenance, no doubt about that.

palinek:
right, the one in #118 is better

yan12125:
Thanks! #118 is much cleaner. Closing in favor of the new PR.

PR #118, opened 2017-04-19
Fixes a FTBFS with CMake v3.8

luis-pereira:
CMake v3.8 AUTOUIC changed the directory where the file is generated to:
<CMAKE_CURRENT_BINARY_DIR>/<TARGETNAME>_autogen/include. The changes makes
us fail.
The change is actually just an sympton and not a root cause. We shouldn't
include a ui generated file in public header.

Closes https://github.com/lxde/lxqt/issues/1277

### DIFF
--- a/liblxqt-config-cursor/selectwnd.cpp
+++ b/liblxqt-config-cursor/selectwnd.cpp
@@ -13,6 +13,7 @@
 #include <QDebug>
 
 #include "selectwnd.h"
+#include "ui_selectwnd.h"
 
 #include <QKeyEvent>
 #include <QMessageBox>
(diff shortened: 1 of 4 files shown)


palinek:
GTM

liblxqt-config-cursor/selectwnd.cpp
```diff
@@ -37,22 +38,24 @@
 
 #define HOME_ICON_DIR QDir::homePath() + "/.icons"
 
-SelectWnd::SelectWnd(LXQt::Settings* settings, QWidget *parent) : QWidget(parent), mSettings(settings)
+SelectWnd::SelectWnd(LXQt::Settings* settings, QWidget *parent)
+    : QWidget(parent),
+      mSettings(settings),
+      ui(new Ui::SelectWnd)
```
palinek: Hm...actually, the `Ui::SelectWnd` object gets never destroyed. We should use a smart pointer here.

yan12125:
As now ```"${CMAKE_CURRENT_BINARY_DIR}/../liblxqt-config-cursor"``` is not needed, how about removing them?

liblxqt-config-cursor/selectwnd.cpp
luis-pereira: Upsss
Forgot to amend the commit. I simply had this:
``` diff
diff --git a/liblxqt-config-cursor/selectwnd.cpp b/liblxqt-config-cursor/selectwnd.cpp
index ccd6312..8a1fa18 100644
--- a/liblxqt-config-cursor/selectwnd.cpp
+++ b/liblxqt-config-cursor/selectwnd.cpp
@@ -83,6 +83,7 @@ SelectWnd::SelectWnd(LXQt::Settings* settings, QWidget *parent)
 
 SelectWnd::~SelectWnd()
 {
+    delete ui;
 }
```

luis-pereira:
@yan12125 That I didn't see. It probably can. Will test it.

yan12125:
I guess this is ready to merge? @palinek

palinek:
Yes, ship it, please.

PR #120, opened 2017-05-05
lxqt-config-file-associations: Sets the initial focus

luis-pereira:
Sets the initial focus to the search text box.


tsujan:
GTM.

agaida:
GTM

PR #127, opened 2017-08-12
lxqt-config-locale: corrected/reworded Save Changed Settings dialog

krukai:
I stumbled over a grammatical error in lxqt-config-locale's changed settings dialog.

I took the liberty of not only removing the irritating misplaced space, but also rewording the dialog slightly.


palinek:
GTM

PR #128, opened 2017-09-16
Update monitorsettingsdialog.ui

stefonarch:
Expanding right panel to avoid truncate tabs.
monitor1
after:
 
schermata-16-21-38-22

### DIFF
--- a/lxqt-config-monitor/monitorsettingsdialog.ui
+++ b/lxqt-config-monitor/monitorsettingsdialog.ui
@@ -56,7 +56,7 @@
       </property>
       <property name="minimumSize">
        <size>
-        <width>300</width>
+        <width>350</width>
         <height>0</height>
        </size>
       </property>


tsujan:
My 2 cents: This depends on the font size and other factors. If tabbar scroll buttons are to be avoided, that could (and should) be done in the code.

stefonarch:
Sure it will not prevent scroll buttons in all situations, but how to do it in the code is far out of my range...

tsujan:
The dialog may have enough space with other styles:

monitor_settings

In general, it isn't easy to avoid scroll buttons/bars. In this case, it might be possible with addition code. If only I had more time...

Anyhow, it can be made a little wider. So, GTM.

tsujan:
Yet another reason for a "GTM":

If scroll buttons/bars are avoided at all costs, the dialog might be too wide with, for example, the Klingon language.

agaida:
we coul expand the width a little bit, just to be sure in the most cases

PR #135, opened 2017-12-08
Update Catalan translation

rbuj:


palinek:
landed in 6123e40d, thanks

PR #137, opened 2018-01-03
Drop Qt foreach

luis-pereira:
Replaced by the ranged-based for loop.
Qt 5.7.1 is required. We use qAsConst().


lxqt-config-monitor/monitor.cpp
```diff
@@ -28,20 +28,20 @@ bool MonitorSavedSettings::operator==(const MonitorSavedSettings &obj)
...
 }
 
-void saveMonitorSettings(QSettings & settings, QList<MonitorSettings> monitors)
+void saveMonitorSettings(QSettings & settings, const QList<MonitorSettings> monitors)
 {
     settings.remove("settings");
     settings.beginWriteArray("settings");
     int i = 0;
-    Q_FOREACH(MonitorSettings monitor, monitors)
+    for(const MonitorSettings& monitor : qAsConst(monitors))
```
palinek: `monitors` is already const...

lxqt-config-brightness/main.cpp
```diff
@@ -54,16 +54,17 @@ int main(int argn, char* argv[])
...
-        QList<MonitorInfo> monitors = brightness->getMonitorsInfo();
+        const QList<MonitorInfo> monitors = brightness->getMonitorsInfo();
         QList<MonitorInfo> monitorsChanged;
         float sign = parser.isSet(decreaseOption)?-1.0:1.0;
         double brightness_value = parser.value(setOption).toFloat();
         brightness_value = qMin( qMax(brightness_value, 0.0), 100.0 ) / 100.0;
         if(!parser.value(setOption).isEmpty())
             sign = 0.0;
-        foreach(MonitorInfo monitor, monitors)
+        for (auto i = monitors.cbegin(); i != monitors.cend(); ++i)
```
palinek: Just curious... why aren't you using the range-for loop here?

palinek:
GTM

lxqt-config-monitor/monitor.cpp
luis-pereira: Thanks

lxqt-config-brightness/main.cpp
luis-pereira: It's an artifact. An experiment to trigger a container detachment.
Actually I inadvertently pushed lxqt-config without a final revision.

luis-pereira:
@palinek Updated.

palinek:
As I said before... GTM

PR #138, opened 2018-01-04
Prevent possible container detachments

luis-pereira:
Found them while performing the drop foreach task.
They hurt performance.


tsujan:
GTM.

PR #141, opened 2018-02-26
Fixed two typos

agaida:


tsujan:
I hadn't read it before. The English sounds wrong to me. Some suggestions:
```
With this application you can set and configure the Qt/LXQt gui appearance of programs.

In Linux, GUI applications depend on their respective subsystems; in LXQt, the overal aspect of
many apps relies on the Qt framework gui. The \fBlxqt-config-appearance\fR application provides
all aspects of setting the \fBLXQt\fR desktop in session settings related programs.
```

tsujan:
"in session settings related programs" sounds a little German ;)

agaida:
to be true - both typos was a lintian finding :P

tsujan:
This is also wrong -- conceptually too:
```
All remaining mayor aspect for lxqt appearance relies over qt4 config application (\fBqtconfig\-qt4\fR), due
the \fBLXQt\fR DE are based on Qt framework. 
```
I suggest:
```
All the remaining major aspects of LXQt appearance rely on the active Qt widget style because the \fBLXQt\fR DE is based on the Qt framework. 
```

Also this:
```
These components perform similar actions to those available in other desktop
environments, and their name is self-descriptive.  They are usually not launched
by hand but automatically, when choosing a \fBLXQt\-qt\fR session in the Display
Manager.
```
My suggestion:
```
These components perform actions similar to those available in other desktop
environments, and their names are self-descriptive.  They are usually not launched
manually but automatically, by choosing an \fBLXQt\-qt\fR session in the Display
Manager.
```
This was as far as I had time -- busy with coding right now.

tsujan:
> Oh, and "QT4" should be "Qt 4"

Actually, Qt4 should be removed, like I suggested above.

@tsimonq2 If you're a native English speaker, would you please correct it, considering what I mentioned about Qt4 above? I think none of us are native speakers.

tsujan:
@tsimonq2  I ask that because if I didn't know about LXQt and just saw that manual, I'd probably never tried it :)

tsimonq2:
Yes, I am a native English speaker; put on my TODO list.

tsujan:
Made the manual readable at https://github.com/lxde/lxqt-config/pull/141/commits/8a7fdca4f6d46dcd096bc8ccece8aa7f13464e33. I think this is good for now -- Probable typos could be found and fixed later.

PR #142, opened 2018-03-03
Fix combo boxes

tsujan:
By sorting their items and adding a comfortable filtering to them (plus a clear button).

Feel free to modify the cmake file if needed.


agaida:
GTM

PR #143, opened 2018-03-18
Backlight control updated.

selairi:


tsujan:
On changing the slider position, if "No" is selected in the prompt dialog, another prompt dialog will also be shown after the first one is closed. This issue might have existed before your work -- I don't know and don't have time to reverse the changes and test.

tsujan:
Another usability case: Wouldn't it be better not to allow backlights < 10% (or 5%) in order to prevent a blank screen by mistake? Again, this may not be directly related to this work -- I'd always changed backlight manually before.

selairi:
> On changing the slider position, if "No" is selected in the prompt dialog,

Fixed.

>  Wouldn't it be better not to allow backlights < 10% (or 5%) in order to prevent a blank screen by mistake?

I have set the minimum backlight to 2%. In my computer it isn't a dark screen, but we should test in other hardware to set the best minimum. In my hardware 5% of back light could be too much brightness in some cases.
If you set the backlight to 0%, it could blank the screen in the most of computers, but I have found computers that set 0% of the backlight is not a blank screen. 
Maybe, a value between 2% and 5% should be enough.

tsujan:
Thanks! Since we can't know about all screens, 2% seems good to me too.

agaida:
GTM

PR #147, opened 2018-04-19
lxqt-config-input: Adds the ability to set the initial page

luis-pereira:
The initial page can be set using the command line options.
It's allows other LXQt modules to programmatically select the initial page
shown to the user.


agaida:
GTM

PR #149, opened 2018-05-09
lxqt-config-brightness: Sets minimum backlight value to 0

luis-pereira:
According to the kernel documentation, backlight vValues are between 0 and
max_brightness.

Reference:
https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-class-backlight

### DIFF
--- a/lxqt-config-brightness/brightnesssettings.cpp
+++ b/lxqt-config-brightness/brightnesssettings.cpp
@@ -33,7 +33,7 @@ BrightnessSettings::BrightnessSettings(QWidget *parent):QDialog(parent)
     ui->backlightSlider->setEnabled(mBacklight->isBacklightAvailable() || mBacklight->isBacklightOff());
     if(mBacklight->isBacklightAvailable()) {
         ui->backlightSlider->setMaximum(mBacklight->getMaxBacklight());
-        ui->backlightSlider->setMinimum((float)(mBacklight->getMaxBacklight())*0.02);
+        ui->backlightSlider->setMinimum(0);
         ui->backlightSlider->setValue(mLastBacklightValue = mBacklight->getBacklight());
         connect(ui->backlightSlider, &QSlider::valueChanged, this, &BrightnessSettings::setBacklight);
     }


luis-pereira:
Yes, 0 doesn't means black screen.
The keyboard keys actually set it to 0.
If you use a laptop check it using the keyboard keys.


--
Cumprimentos
Regards

A quarta, 9/05/2018, 21:23, Alf Gaida <notifications@github.com> escreveu:

tsujan:
That `(float)(mBacklight->getMaxBacklight())*0.02` was included intentionally to prevent a black screen. Even if `0` doesn't result in a black screen on some laptops, it's safer not to use it. On the other hand, who can work with a such a dark screen?

tsujan:
IMO, 2% of the max brightness is a reasonable value.

luis-pereira:
@agaida  @tsujan `/sys/class/backlight/<device>/brightness` takes an `int` not a `float`.  What's the point of multiplying it by 0.02 ?
If black screen was a concern the solution is to set the minimum to 1 not multiplying by some random value. 

Can you guys, pls, set the backlight brightness no the minimum (using keys, or under KDE, XFCE, etc) and check what's the content of `/sys/class/backlight/acpi_video0/actual_brightness`.

p.s. We are talking about backlight, not `xrandr` brightness. That's another thing.

tsujan:
> takes an int not a float

You're right -- I'd forgotten that -- which means that line is wrong. However, it can be fixed easily and if you set it to zero, I'll have to change it again when fixing https://github.com/lxqt/lxqt/issues/1479.

tsujan:
The fix can be something like this:
```cpp
ui->backlightSlider->setMinimum(qMax(qRound((qreal)(mBacklight->getMaxBacklight())*0.02)), 1)
```
Which isn't so different from the current line.

luis-pereira:
@tsujan 0 is a safe value in this context. It's the minimum the hardware allows. It doesn't mean black screen. It most systems it doesn't. Even if it does, whe have the timer functionality. It will revert back. 

Who do we think we are to impose arbitrary limits to the user ?

tsujan:
> Even if it does, whe have the timer functionality

That's what I want to remove. KDE doesn't have it and it's a hassle.

> It doesn't mean black screen.

On my laptop, it isn't a black screen but very dark -- I don't know others could see the screen like I can -- and we can't be sure about all laptops, while 2% is reasonable.

tsujan:
> Who do we think we are to impose arbitrary limits to the user ?

Strange sentence! This matter doesn't have anything to do with anyone's personality.

luis-pereira:
> Which isn't so different from the current line.

``` cpp
ui->backlightSlider->setMinimum(1)
```
to do that. is much simpler.

tsujan:
>  is much simpler.

Simpler but not efficient. On my laptop, 1 doesn't have any difference from 0 but 2% does: 2% is dim and visible but, of course, not enough for real work. I vaguely remember that I proposed 1% but agreed with 2.

luis-pereira:
> That's what I want to remove. KDE doesn't have it and it's a hassle.

KDE allows setting the value from 0 to max_brightness. Other DE's do the same. And they are right IMO.
What KDE doesn't have is the dial thing. That should be removed IMO.
Remove the dial thing and the timer it's not needed.

> On my laptop, it isn't a black screen but very dark -- I don't know others could see the screen like I can -- and we can't be sure about all laptops, while 2% is reasonable.

That makes sense for the dial thing not for the backlight.

> Strange sentence! This matter doesn't have anything to do with anyone's personality.

I,m not talking about personality, just let the user choose.

tsujan:
Numbers: for me 937 × 2% = 19, which is way different from 1 but very dim + visible. On my other laptop, it has a different value but still very dim.

luis-pereira:
> Simpler but not efficient. On my laptop, 1 doesn't have any difference from 0 but 2% does: 2% is dim and visible but, of course, not enough for real work. I vaguely remember that I proposed 1% but agreed with 2.

There is no percentage. The backlight brightness is **not**  a continuous variable.
It's a discrete variable with [0, max_brightness] range.

The kernel guys made it very simple.

luis-pereira:
> Numbers: for me 937 × 2% = 19, which is way different from 1 but very dim + visible. On my other laptop, it has a different value but still very dim.

To me here is [0,7]. What we need to do is define a step value. From working in other systems is what KDE does, I think.

tsujan:
> There is no percentage.  The backlight brightness is not a continuous variable.

There's always a percentage. 50% of 7 is 3. Actually the current line is correct; when I said it was wrong, I just meant `qRound` and `1`, which aren't important at all.

luis-pereira:
I will implement the step logic and then we can discuss seeing the actual code.

tsujan:
A 10 step slider would be very poor for all laptops I've seen. And max(20% × 7, 1) == 1.

tsujan:
And steps have nothing to do with this.

selairi:
> What KDE doesn't have is the dial thing. That should be removed IMO.
> Remove the dial thing and the timer it's not needed.

The dial changes brightness (by software). The slider chages the backlight.
In a desktop computer you can not change the backlight, but you can change
the brightness.

I have found some screen devices too much brightness and it is not good for
my eyes. In those screens, you can not set a lower brightness level, but
you can set it by software. The dial can be useful and it not should be
removed IMO.

I am thinking in add some QLabels explaining the difference in
lxqt-config-brightness.

On the other hand, the backlight minimum is a really complex issue. I have
found laptops that 0 value in /sys/class/backlight/<device>/brightness turns
off the screen, but there are others that 0 value is the minimum backlight
level (and it doesn't turn off the screen).

If you laptop turns off the screen at 0 value of backlight, Systemd will
remember that backlight level and the screen will be turned off in the next
reboot. I think that is not a good idea.

I think that other DEs set the minimum backlight level as we do.

Maybe, the minimum level could be tested on lxqt-config-brightness at first
run. We could set the backlight to 0 for 2 or 3 seconds and check if
/sys/class/backlight/<device>/*bl_power* is set to 4 (the screen has turned
off).

What do you think?

2018-05-10 0:47 GMT+02:00 tsujan <notifications@github.com>:

selairi:
> The only thing that works is to define a number of steps for that slider.

I humbly do not think so. The response of the eye to light is not a linear
function. The eye is more sensitive to minor brightness. If steps are
inserted, we will lose this control at low backlight levels.

2018-05-10 11:48 GMT+02:00 P L. Lucas <selairi@gmail.com>:

luis-pereira:
> A 10 step slider would be very poor for all laptops I've seen. And max(20% × 7, 1) == 1.

10 was just to simplify the discussion, not an actual propose.The definition of the step is another discussion.
The code has 2% X 7 = 0,14.

luis-pereira:
> And steps have nothing to do with this.

When one uses a slider (a step based control), it has.

luis-pereira:
@selairi There are interesting points in your post. Will come back on this very soon.

tsujan:
> Maybe, the minimum level could be tested on lxqt-config-brightness at first run. We could set the...

I think other DEs take it easier -- sometimes too much ;) After testing on 3 laptops, I set it to 5% at https://github.com/lxqt/lxqt-config/pull/150 . 2% seemed too dark and 0 was not an option.

The QSlider issue I mentioned [elsewhere](https://github.com/lxqt/lxqt/issues/1485) is about how Qt sets the max and min. It can also be seen as a property, in contrast to issue: QSlider may set the max value 1 unit more than what it should be, so I took that into account in https://github.com/lxqt/liblxqt/pull/138

agaida:
GTM

PR #150, opened 2018-05-10
Remove time-out from backlight slider

tsujan:
Instead, prevent too fast changes to the backlight value by using a `QTimer`.

Also, set the minimum backlight value to 5% of its maximum because 2% is too dark.


tsujan:
NOTE: This should be compiled against https://github.com/lxqt/liblxqt/pull/138 for the max value to be set correctly.

agaida:
bah damn webinterface

PR #151, opened 2018-05-11
Qt 5.11 beta: timeoutdialog.cpp: include QIcon, QStyle explicitely

MaxKrummenacher:
5.11 Qt headers do not indirectly include QIcon, QStyle resulting in compile
time errors like:

| timeoutdialog.cpp:32:25: error: invalid use of incomplete type 'class QStyle'

Signed-off-by: Max Krummenacher <max.krummenacher@toradex.com>


agaida:
refs: lxqt/lxqt/issues/1487

palinek:
GTM

PR #154, opened 2018-05-31
up/down to Up/Down

scootergrisen:


agaida:
Ha - the first string change to be handled by Weblate :P  GTM

PR #227, opened 2018-07-10
Remove unnecessary semicolon

rikardfalkeborn:
Fixes a compiler warning with Gcc:

    In file included from lxqt-config/src/main.cpp:30:
        lxqt-config/src/mainwindow.h:71:2: warning: extra ‘;’ [-Wpedantic]


agaida:
Thanks, nice catch

PR #280, opened 2018-09-11
Small GUI fixes

tsujan:
Among other things:

 * `QStyle::State_HasFocus` is only for focused viewports.
 * A single label line should be wrapped without internal line-ends.


agaida:
gtm

PR #281, opened 2018-09-15
Delete untranslated lxqt-config-input_zh_HK.ts

scootergrisen:
Perhaps it should be deleted on Weblate to?


agaida:
i guess you are right

PR #322, opened 2018-11-09
Monitor name is added to output.

selairi:
The name of the screen is added to output name. In this picture the HDMI output shows the name of screen:
monitor-name
Users can identify their screens easier.


selairi:
OK. I have just added the brackets.

PR #357, opened 2018-12-11
Fixed touchpad settings

tsujan:
(1) The name is changed to "Mouse and Touchpad". The "Mouse" section should be either removed or merged with it in future.

(2) The GUI wasn't updated on changing the device. That's fixed by this patch.

(3) The signals are changed so that the slots are called only with user interaction, and not on device change. In the case of the spinbox, `blockSignals()` is called when the value is changed by the code (on device change). The slots are also modified accordingly.

(4) The variable `curDevice` is removed.


tsujan:
@yan12125 Please ignore `touchpadconfig.ui` (just the layout is changed) and check the other changes!

Issue #366, opened 2018-12-16
X11_Xinput_LIB not found


jubalh:
Latest git gives me:

```
[   92s] -- Found X11: /usr/lib64/libX11.so
[   92s] -- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.2") 
[   92s] -- Checking for modules 'xcb;xcb'
[   92s] --   Found xcb, version 1.13
[   92s] --   Found xcb, version 1.13
[   92s] -- Found XCB: /usr/lib64/libxcb.so;/usr/lib64/libxcb.so  
[   92s] -- Found ZLIB: /usr/lib64/libz.so (found version "1.2.11") 
[   92s] -- Checking for module 'xorg-libinput'
[   93s] --   Found xorg-libinput, version 0.28.1
[   93s] -- Checking for modules 'xcb;xcb;xcb-randr'
[   93s] --   Found xcb, version 1.13
[   93s] --   Found xcb, version 1.13
[   93s] --   Found xcb-randr, version 1.13
[   93s] -- Found XCB: /usr/lib64/libxcb.so;/usr/lib64/libxcb.so;/usr/lib64/libxcb-randr.so  
[   93s] CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
[   93s] Please set them or make sure they are set and tested correctly in the CMake files:
[   93s] X11_Xinput_LIB (ADVANCED)
```

`xorg-libinput` devel package is installed. @yan12125 shouldn't `X11_Xinput_LIB` be set automatically?

yan12125:
Hmm, have never investigated how X11_Xinput_LIB works. I guess /usr/lib/pkgconfig/xorg-libinput.pc is the necessary file. It is provided by xf86-input-libinput on Arch Linux. BTW, I remeber you're using OpenSUSE?

yan12125:
Oops turns out I was wrong. Relevant files are /usr/lib/libXi.so and /usr/include/X11/extensions/XInput.h, which are provided by libxi on Arch.

jubalh:
Right, that helps. Thanks @yan12125!

Can I add `pkg_check_modules(XORG_XI xi)` to CMake in -input subdir?

jubalh:
libudev is also needed, shouldnt we also check for this?

yan12125:
> Can I add `pkg_check_modules(XORG_XI xi)` to CMake in -input subdir?

[...]

PR #368, opened 2018-12-16
Check fof xi and libudev

jubalh:
We require libXi.so and libudev.h.
See https://github.com/lxqt/lxqt-config/issues/366


agaida:
apply and build fine - GTM

yan12125:
Looks good and build fine here, too.

Issue #358, opened 2018-12-12
Package 'xorg-libinput' Required


Chiitoo:
After ee7b0fee [1], 'xorg-libinput' seems to be mandatory, which doesn't make sense on devices not making use of it, and adds unused package for them.

Whould it be possible to have it optional instead?

Thanks!

1. https://github.com/lxqt/lxqt-config/commit/ee7b0fee1b14913248b864ae8d37e7c24df4af75#diff-31004112b81349ff2016dc1a2eb3d5f1R5

PR #374, opened 2018-12-29
build: make touchpad settings/dependencies optional (#358)

Chiitoo:
Me and my options...  :]

This might not be perfect, or even completely correct in its everything, but I figured I'd ask if anything like this would be acceptable?

Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,7 @@ option(WITH_FILE_ASSOCIATIONS "Build the 'lxqt-config-file-associations'" ON)
 option(WITH_INPUT "Build the 'lxqt-config-input'" ON)
 option(WITH_LOCALE "Build the 'lxqt-config-locale'" ON)
 option(WITH_MONITOR "Build the 'lxqt-config-monitor'" ON)
+option(WITH_TOUCHPAD "Build touchpad settings" ON)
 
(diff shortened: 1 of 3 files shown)


yan12125:
Adding such an option is OK for me.

palinek:
Seems OK...but isn't creating a config.h file an overkill for one define?

Chiitoo:
> Seems OK...but isn't creating a config.h file an overkill for one define?

It's just the only method of doing it that I have seen (or remember), so it certainly could be!  :]

I'll go look around more.  Thanks!

yan12125:
Maybe `add_definitions` can get to a similar goal?

By the way, I prefer to put `option` commands in the top level CMakeLists.txt. As a user I look at that file first for possible options.

Chiitoo:
Changed things up, and it seems to work for me at least.

I'm not too sure about the way I sorted and grouped things up, so any opinions and/or suggestions are more than welcome.  :]

yan12125:
Done! Thanks for the enhancement!

> It seems to be superseded by alternatives such as 'add_compile_definitions()', but that's not until CMake 3.12.something (looks like the minimum version supported right now for LXQt is 3.1.0).

I'd say requiring 3.12+ is too aggressive. Let's keep using add_definitions() until there's another problem.

PR #394, opened 2019-01-30
Apply button for appearance and input config dialogs

tsujan:
Closes https://github.com/lxqt/lxqt/issues/1625

Depends on https://github.com/lxqt/liblxqt/pull/202

Since the appearance and input config dialogs had cursor themes in common (which is good, IMO), the patch should have covered both; hence its big size.

Its logic is simple though: When a setting is changed, don't apply it immediately but enable the Apply button for the user (as in KDE). If several settings are changed, the Apply button will save and apply all of them when clicked.

NOTE: Rebasing this PR would be VERY boring to me. Please don't torture me!


tsujan:
Oh, trailing spaces are removed by FeatherPad in some files; it wasn't intentional but isn't bad either.

PR #460, opened 2019-04-10
Pedantic changes to texts of touchpad tap settings

tsujan:
They were a little confusing before.

### DIFF
--- a/lxqt-config-input/touchpadconfig.ui
+++ b/lxqt-config-input/touchpadconfig.ui
@@ -34,7 +34,7 @@
    <item row="4" column="0" colspan="2">
     <widget class="QCheckBox" name="tappingEnabledCheckBox">
      <property name="text">
-      <string>Single click to activate items</string>
+      <string>Tap to click</string>
      </property>
     </widget>
    </item>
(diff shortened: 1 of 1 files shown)


yan12125:
Thanks! Maybe "Tap twice and drag" is even clearer?

tsujan:
> Maybe "Tap twice and drag" is even clearer?

The second movement isn't a tap; only the fist one is.

tsujan:
> as tap twice and drag (3 actions in total) also works

Strange! Not here. If I tap twice on a GitHub image link, it'll be opened before I can drag it. I guess touchpads are different. Tapping and then dragging doesn't open the link but drags the image, as expected.

yan12125:
> I guess touchpads are different.

Maybe. I use bcm5974 on MacBookAir7,2. Tap twice and drag works for folders and files in pcmanfm-qt :)

tsujan:
Maybe we don't mean the same thing.

One tapping = Hitting + moving the finger away from touchpad. The second time, the finger hits the pad and slides on it (no tapping). Hence, it's one tapping and one dragging (= hit+slide).

If not, our touchpads are really different.

yan12125:
We do have different touchpads :)

yan12125:
Oops, I almost forgot this PR. Merged as it's simple and just correct.

PR #465, opened 2019-05-14
Avoid char raw strings literals memory allocations

luis-pereira:
Done by CLAZY_FIXIT="fix-fromCharPtrAllocations".


tsujan:
This should be merged before another PR is made. I'll check it.

Issue #508, opened 2019-07-20
Regression: lxqt-config-appearance overwrites GTK setting even when it should not


tsujan:
Because of https://github.com/lxqt/lxqt-config/commit/e265219437568c58485242c5083325670962703b, when the box "Set GTK themes" isn't checked, clicking the Apply button overwrites GTK settings.

lxqt-config-appearance should respect the user's choice and do nothing with GTK settings if that box isn't checked -- as it did before.

PR #509, opened 2019-07-20
Apply GTK settings only if the user chooses that

tsujan:
Fixes https://github.com/lxqt/lxqt-config/issues/508

Changing GTK settings is optional in the GUI; so, they shouldn't be overwritten if the user doesn't want that.

As to why changing GTK settings should remain optional, suffice it to say that the user may have a GTK-based DE installed alongside LXQt, although that isn't the only reason.


agaida:
GTM

Issue #511, opened 2019-07-28
lxqt-config-brightness: Use native backlight control instead of fake brightness controll when available


lupinix:
If native backlight control is available, a slider in lxqt-config-control allows to use it instead of the "faked" brightness control. However, this is not the case when we run lxqt-config-brightness with -i or -d to decrease/increase brightness using command line. I request to add that functionality here, too.

The main reason: lxqt-config-brightness -d/-i is used for the keyboard shortcuts for brightness controll. As a result, using the shortcuts, the user always ends up with the faked brightness control, even when the native one is available.

selairi:
Yes. It should be fixed. Lxqt backlight control is only used in GUIn only.
I would try to fixed.

lupinix:
Thanks for investigating :)

selairi:
There is a new pull request,  #521, that solves this problem. It must be accepted in master branch. Be patient.

shlyakpavel:
The PR above is merged. Can we close the issue?

agaida:
We can - the only thing left is to fix the ~~fucking ugly~~ not so usable interface.

tsujan:
> the only thing left is to fix the not so usable interface.

It needs a little more love:

![brightness](https://user-images.githubusercontent.com/981076/63965802-304dbf80-caaf-11e9-9b73-b7bbd8eb95ef.png)

Shouldn't be hard. First remover that dial, then add ticks and numbers to the slider; problem solved.

agaida:
right - i'm only mentioned it

selairi:
I think that this issue should be closed.

tsujan:
I'll close it after https://github.com/lxqt/lxqt-config/pull/612 is merged. Thanks for the reminder!

tsujan:
Closed in https://github.com/lxqt/lxqt-config/pull/612

PR #521, opened 2019-08-05
Use backlight driver in command line.

selairi:
Only XRandr driver was used in command line options. If backlight driver is available, it will be used in order to change backlight.

See #511


lupinix:
Thanks! I'll try it out until friday.

lupinix:
Works as expected on my ThinkPad X230! Thank you very much!

PR #526, opened 2019-08-23
lxqt-config-monitor: Drag and drop align monitors method

selairi:
A new method to drag and drop monitors has been written for lxqt-config-monitor. Monitors will be aligned if the screens near the corners.

This method also solves a bug casting from float to int:
QGraphicsView uses float to store the elements positions. MonitorPicture uses int. The cast from float to int will sometimes lose one or two pixels and the screens are not connected if you enable the "Keep monitors attached" checkbox. Then, the users must to change the positions with GUI's spinboxes.

See [#1738.](https://github.com/lxqt/lxqt/issues/1738)


tsujan:
I confirm that this patch fixes the problematic behavior of DND alignment.

PR #571, opened 2019-11-30
Add option to scale widgets to double size

jdoxey:
Selecting this option is a shortcut which adds/sets the following environment variables (under Session Settings > Environment (Advanced)),
- QT_SCALE_FACTOR=2
- QT_AUTO_SCREEN_SCALE_FACTOR=0 (to stop HiDPI Qt apps like Qt Designer from scaling twice)
- GDK_SCALE=2
- XCURSOR_SIZE=32
as described in this thread https://forum.lxqt.org/t/detailed-guide-to-enable-high-dpi-scaling-on-lxqt/507

### DIFF
--- a/lxqt-config-appearance/main.cpp
+++ b/lxqt-config-appearance/main.cpp
@@ -67,7 +67,7 @@ int main (int argc, char **argv)
     QSettings& qtSettings = *settings; // use lxqt config file for Qt settings in Qt5.
 
     /*** Widget Style ***/
-    StyleConfig* stylePage = new StyleConfig(settings, &qtSettings, &mConfigAppearanceSettings, configOtherToolKits, dialog);
+    StyleConfig* stylePage = new StyleConfig(settings, &qtSettings, &mConfigAppearanceSettings, configOtherToolKits, sessionSettings, dialog);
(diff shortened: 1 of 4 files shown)


tsujan:
This is a nice idea that needs more work.

IMHO, there are 4 problems with this approach. From important to small:

 1. There are four EVs but only one option: to be or not to be.
 2. The values -- especially that of `QT_SCALE_FACTOR` -- aren't chosen by the user. Appropriate sin-boxes/combo-boxes are missing.
 3. Widget style is not the right place because this isn't about styling. (All other items *are* about styling.)
 4. In the explanation, it's supposed that Openbox is LXQt's WM.

jdoxey:
Regarding points 1 and 2: Yes, maybe the values should be editable on the same screen, without having to go to Session Settings > Environment.
Regarding point 3: Where do you think an option like this should live? (Eventually it would be nice in the Monitor settings, so you can set it separately for each screen, but that feature will probably present a new set of challenges)

tsujan:
>  Where do you think an option like this should live?

`XCURSOR_SIZE` can go to the Cursor section of LXQt Appearance Configuration, with an appropriate setting of its own. The other three EVs can have a separate "Scaling" section -- again, with appropriate settings, like spin/combo boxes.

That was only a suggestion. I like to know other devs think, especially @selairi, who's the expert in this area.

But I like the idea and think that it should be implemented.

PR #575, opened 2019-12-11
Remove deprecated qSort

luis-pereira:
Use std::sort instead.


tsujan:
Ah, thanks! I'd noticed them but, after a while, forgotten them.

luis-pereira:
They clutter build outputs and deprecated.
There are more qXXXX stuff deprecated. Will push fixes.

PR #608, opened 2020-03-13
Removed old mouse accel/sensitivity options

tsujan:
They had no effect and were confusing. The real options are under "Mouse and Touchpad".

NOTE: Later, we could discuss how to rearrange the dialog's sections but, IMHO, those options should have been removed as soon as possible.

### DIFF
--- a/lxqt-config-input/mouseconfig.cpp
+++ b/lxqt-config-input/mouseconfig.cpp
@@ -59,10 +59,6 @@ MouseConfig::MouseConfig(LXQt::Settings* _settings, QSettings* _qtSettings, QWid
   loadSettings();
   initControls();
 
-  // set_range_stops(ui.mouseAccel, 10);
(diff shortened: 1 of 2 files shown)


tsujan:
> I agree with a Ui refactoring.

I think we'll need to do it after the next release, if it happens soon -- in order to have time to test it.

Chiitoo:
They do have an effect when, I believe, 'xf86-input-libinput' is not installed.

Granted, the effect is a bit... odd, but they /do/ do stuff.  :]

The 'Mouse and Touchpad' settings might not even exist depending on the build configuration, but I understand if this is not something that people want to maintain.

tsujan:
@Chiitoo `libinput` is the "new" way (actually, it isn't new anymore). Those options could be very confusing to most users. I think KDE removed them too.

Chiitoo:
Yes, I have 'libinput' installed, but the controls still work.  They don't do anything if I also have the "old" way installed, which is required for the 'Mouse and Touchpad' settings (which actually seems a bit odd since it's the "old" way).

Issue #404, opened 2018-04-25
Poor user experience in lxqt-config-brightness


AlpacaRotorvator:
I'm running lxqt on a laptop and I found the way the brightness controls work to be very suboptimal. My main gripes with it currently are:

*  There's no telling what each control does. The user has to move them around to get an intuition of their function.

* The dial-style control is very unwieldy on a trackpad, resulting on me swinging it around violently or unintentionally letting go of it while trying to fine tune.

* The way it currently works you can't easily try a setting that alters both controls. You have to move one control, save your configuration, then tweak the other one, save, then perhaps tweak the first one again, save, and so forth. The modal dialog is just too disruptive.

I'm no UX expert, but what I suggest to alleviate those problems is:

* Individual labels for each control.

* Make the control that is currently a dial(brightness?) into a slider like the other one(backlight?).

* This is the trickiest of the three. On one hand the user should be able to move the controls around and immediately see what the screen would look like, on the other hand we want a failsafe in case he accidentally sets his screen too bright/dark, but it shouldn't hassle the user too much mid-tuning like it does now.

One solution I can think off the top of my head would be do away with the modal popup and just have an "Apply/Save" button on the main window. Then have a timer running that resets every time the user changes the value of a control, and if the timer expires before the user clicks the apply button it automatically reverts to the previous value.
 
Not sure if relevant here but, just in case, I'm running LXQt 0.12 on Qt 5.10.1.

selairi:
Hi, I have done some changes in the UI:
![lxqt-config-brightness-ui](https://user-images.githubusercontent.com/1163139/79002647-2fa2fe80-7b51-11ea-9c74-470656bf76ab.png)

What do you think?

tsujan:
@selairi 
Your screenshot is much better than the current GUI.

selairi:
I have added reset, increase and decrease buttons:
![bajarbrillo](https://user-images.githubusercontent.com/1163139/79051635-6fd9ae00-7c31-11ea-8037-bdb7ba3e9ee9.png)
Suggestions?

tsujan:
@selairi
IMO, it's perfect.

BTW, I'll test your last commit at https://github.com/lxqt/lxqt-powermanagement/pull/189 soon.

tsujan:
Closed in https://github.com/lxqt/lxqt-config/commit/3421fb5ad8d4b0de1a200391f4727a691031ebfe

PR #612, opened 2020-04-12
lxqt-config-brightness better UI

selairi:
Some changes are done in lxqt-config-brightness UI:
bajarbrillo

Related to [https://github.com/lxqt/lxqt-config/issues/404](https://github.com/lxqt/lxqt-config/issues/404)

### DIFF
--- a/lxqt-config-brightness/brightnesssettings.cpp
+++ b/lxqt-config-brightness/brightnesssettings.cpp
@@ -36,20 +36,7 @@ BrightnessSettings::BrightnessSettings(QWidget *parent):QDialog(parent)
     ui->backlightSlider->setEnabled(mBacklight->isBacklightAvailable() || mBacklight->isBacklightOff());
     ui->backlightGroupBox->setEnabled(mBacklight->isBacklightAvailable() || mBacklight->isBacklightOff());
     if(mBacklight->isBacklightAvailable()) {
-        // Set the minimum to 5% of the maximum to prevent a black screen
-        int minBacklight = qMax(qRound((qreal)(mBacklight->getMaxBacklight())*0.05), 1);
-        int maxBacklight = mBacklight->getMaxBacklight();
-        int interval = maxBacklight - minBacklight;
-        if(interval <= 100) {
-            ui->backlightSlider->setMaximum(maxBacklight);
-            ui->backlightSlider->setMinimum(minBacklight);
-            ui->backlightSlider->setValue(mBacklight->getBacklight());
-        } else {
-            ui->backlightSlider->setMaximum(100);
-            // Set the minimum to 5% of the maximum to prevent a black screen
-            ui->backlightSlider->setMinimum(5);
-            ui->backlightSlider->setValue( (mBacklight->getBacklight() * 100) / maxBacklight);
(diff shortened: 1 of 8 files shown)


tsujan:
This GUI is nice; no confusion anymore.

Just small details — exactly as you fixed https://github.com/lxqt/lxqt-panel/pull/1385:

 1. Mouse wheel on backlight may be hard to use because the scale isn't in percents (my max brightness is 5273). This was a problem before your PR, I think.
 2. For the same reason, the buttons may not change the *visible* backlight.
 3. Please swap `☼` and `☀` with each other.

luis-pereira:
@selairi trailing whitespaces.
It's most certainly your editor that's introducing them.

selairi:
>     1. Mouse wheel on backlight may be hard to use because the scale isn't in percents (my max brightness is 5273). This was a problem before your PR, I think.

Done.
 
>     2. For the same reason, the buttons may not change the _visible_ backlight.

Done.

>     3. Please swap `☼` and `☀` with each other.

Done.

selairi:
> @selairi trailing whitespaces.
> It's most certainly your editor that's introducing them.

@luis-pereira You are right. I have just enabled the option to remove trailing white spaces in my editor. Thanks.

tsujan:
@selairi 
After compiling your last commit, I saw some issues. There may not be new -- I might haven't checked them before -- but I'm not sure.

 1. Confirmation for brightness isn't reliable here. Sometimes it works, sometimes not. For example, if I set the brightness to maximum and don't confirm the change, the slider will go back to its previous position but the brightness won't change: it will remain at its maximum value. It'll be corrected If I change it by grabbing the slider.
 2. Reset isn't reliable for brightness either. For example, after I set the brightness to maximum and confirm the change, if I press the reset button, the slider goes back to its previous position but the brightness doesn't change. Again, grabbing the slider corrects the problem. 
 3. There is no confirmation for backlight.

Sorry if my report isn't clear enough. Tell me if you need more info.

tsujan:
@selairi 
Would you please give priority to this? I want to release lxqt-config soon but this PR is too important to be missing from the next version.

selairi:
@tsujan Fixed!

tsujan:
@selairi Thanks. Will test it later.

It'll go to the next version -- lxqt-config is already released.

tsujan:
@selairi 
With the last change you made, it's fixed for my laptop screen but not for the external monitor: when I move the brightness slider (not the backlight slider) from the middle to the right and do NOT confirm the changes, the slider goes back to its previous position but the high brightness of my external monitor doesn't change.

For the laptop monitor, both brightness and backlight sliders are OK. I'm not sure what the difference is.

selairi:
@tsujan 
> With the last change you made, it's fixed for my laptop screen but not for the external monitor: 

I have tested in my external monitor, but it works properly. I have done some changes in the brightness control. Now, it checks monitors by id.

tsujan:
@selairi 
The problem disappeared after a reboot.

Please reverse your last commit (`if(monitor.id() == (int)output && monitor.name() == name)` had never caused a problem), squash and merge!

PR #621, opened 2020-05-03
lxqt-config-input: add a knob to enable/disable a mouse/touchpad device

yan12125:


tsujan:
GTM.

I'm worried about one thing though:

Suppose that the user has disabled his/her touchpad and then, he disables his mouse and clicks Apply confidently. OK, he still has the keyboard, with which he could enable the mouse again but (1) The tab order is a mess in the GUI and (2) even if we fix the tab order (which we should), the user may not know about tab or may have difficulty in using keyboard. As a result, we'll get bug reports like, "Mouse is disabled under LXQt…".

Adding a visible shortcut might solve the problem only partially because the user may not know how to reach this section with keyboard.

@yan12125 
Is it possible to add "Disable touchpad when mouse is plugged in" instead, like in KDE?

yan12125:
> Disable touchpad when mouse is plugged in

Sounds like a daemon is needed for monitoring events. I'll try to see how to achieve that.

tsujan:
Another possibility may be disabling the checkbox in one device if it's checked for another existing device but various scenarios will become possible (e.g., if the fist device disappears for whatever reason, we'll have no input other than keyboard ).

yan12125:
> Another possibility may be disabling the checkbox in one device if it's checked for another existing device but various scenarios will become possible (e.g., if the fist device disappears for whatever reason, we'll have no input other than keyboard ).

Sounds complex and error-prone :/

---

KDE achieves "Disable touchpad when mouse is plugged in" with a daemon [1] listening to X11 events [2]. Udev events might be better in terms of performance.

Compatibility concerns:
* I don't want more X11-specific codes. Listening to X11 events is not the way.
* udev is a dependency of libinput [3], so it should be availble as long as this dialog runs. As a side note, udev is supported on FreeBSD [4], experimental on NetBSD [5], and appears not working on OpenBSD as of late 2019 [6].

[1] https://github.com/KDE/plasma-desktop/blob/v5.18.5/kcms/touchpad/src/kded/kded.cpp
[2] https://github.com/KDE/plasma-desktop/blob/v5.18.5/kcms/touchpad/src/backends/x11/xlibnotifications.cpp
[3] https://github.com/wayland-project/libinput/blob/1.15.5/meson.build#L125
[4] https://www.freshports.org/devel/libudev-devd
[5] https://pkgsrc.se/wip/udev
[6] https://news.ycombinator.com/item?id=21285177

tsujan:
> I don't want more X11-specific codes. Listening to X11 events is not the way.

I agree completely.

YerCodeSmellsFunny:
My touchpad produces quite a bit of noise, so I would like to disable it.  Instead, I set the acceleration to the lowest value (most negative, that is: -1.0)  and it reduces the noise on my mouse cursor. It would be nice to have a way to disable this completely through the UI, or at least the CLI.  If there is a way to do this, the help file doesn't indicate it.   Thanks for reading my "report". :)

tsujan:
> If there is a way to do this, the help file doesn't indicate it.

Of course, there is but, because it's independent of DEs and ditros, there's no reason to add an LXQt documentation for it. Find your touchpad ID from `xinput list` and disable it with `xinput set-prop your_tp_id "Device Enabled" 0`.

YerCodeSmellsFunny:
> `xinput set-prop your_tp_id "Device Enabled" 0`.
Thank you so much tsujan!  I'm feeling sane again already.  :) 
In case it shows up  as another ID to help others with the same problem,  I made the fix generic in this fashion:

xinput set-prop `xinput list | grep ouchpad |   cut -d '=' -f 2 | cut -c 1-2`  "Device Enabled" 0

Cheers

yan12125:
I don't use Linux on my laptop anymore. I will continue to work on this PR if my next laptop is Linux-friendly. That will not happen in near future, though. If anyone is interested, feel free to create another PR and replace this one.

PR #626, opened 2020-05-17
Screen daemon

selairi:
Simple screen daemon. Related to: [https://github.com/lxqt/lxqt/issues/1788](https://github.com/lxqt/lxqt/issues/1788)


yan12125:
Seems `lxqt-monitor/UdevNotifier.*` are the same as `lxqt-session/src/UdevNotifier.*`? If so, it will be better to keep that class in one place. Maybe liblxqt.

tsujan:
I guess they're meant to be removed from `lxqt-session` in another PR. @selairi, am I right?

> it will be better to keep that class in one place. Maybe liblxqt.

I'm not sure. If they had other purposes, yes. But they're only about monitors. So, IMHO, this place is OK. Maybe it's a habit but I'm more comfortable when I find things that are related to each other in one place; in this case, the current place.

yan12125:
Well, UdevNotifier class is still used in lxqt-session for detecting new input devices and loading keyboard settings - https://github.com/lxqt/lxqt-session/blob/784acb78587c72c7f77dbf8878ef4d51bdffb938/lxqt-session/src/sessionapplication.cpp#L84 (maybe that should be moved from lxqt-session to a standalone daemon, too?)

tsujan:
@yan12125 
I missed that. You're right; liblxqt may be the best choice then.

PR #629, opened 2020-05-31
Added an option to change window color

tsujan:
It is in LXQt Appearance Configuration → Widget Style.

Note: For the option to work, lxqt-qtplugin 0.15.1 should be installed; otherwise, it will have no effect.

### DIFF
--- a/lxqt-config-appearance/colorLabel.cpp
+++ b/lxqt-config-appearance/colorLabel.cpp
@@ -43,8 +43,10 @@ void ColorLabel::setColor(const QColor& color)
     stylesheetColor_ = color;
     // ignore translucency
     stylesheetColor_.setAlpha(255);
-    setStyleSheet(QStringLiteral("QLabel{background-color: rgb(%1, %2, %3);}")
-                  .arg(color.red()).arg(color.green()).arg(color.blue()));
+    QString borderColor = qGray(stylesheetColor_.rgb()) < 255 / 2
+                            ? QStringLiteral("white") : QStringLiteral("black");
+    setStyleSheet(QStringLiteral("QLabel{background-color: rgb(%1, %2, %3); border: 1px solid %4;}}")
+                  .arg(color.red()).arg(color.green()).arg(color.blue()).arg(borderColor));
 }
 
 QColor ColorLabel::getColor() const
--- a/lxqt-config-appearance/colorLabel.cpp
+++ b/lxqt-config-appearance/colorLabel.cpp
@@ -46,7 +46,7 @@ void ColorLabel::setColor(const QColor& color)
     QString borderColor = qGray(stylesheetColor_.rgb()) < 255 / 2
                             ? QStringLiteral("white") : QStringLiteral("black");
     setStyleSheet(QStringLiteral("QLabel{background-color: rgb(%1, %2, %3); border: 1px solid %4;}")
-                  .arg(color.red()).arg(color.green()).arg(color.blue()).arg(borderColor));
+                  .arg(QString::number(color.red()), QString::number(color.green()), QString::number(color.blue()), borderColor));
 }
 
 QColor ColorLabel::getColor() const
--- a/lxqt-config-appearance/colorLabel.cpp
+++ b/lxqt-config-appearance/colorLabel.cpp
@@ -1,0 +1,65 @@
+/* BEGIN_COMMON_COPYRIGHT_HEADER
+ * (c)LGPL2+
+ *
+ * LXQt - a lightweight, Qt based, desktop toolset
+ * https://lxqt.org/
+ *
+ * Copyright: 2020 LXQt team
+ *
+ * This program or library is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General
+ * Public License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA
+ *
+ * END_COMMON_COPYRIGHT_HEADER */
+
+#include "colorLabel.h"
+#include <QColorDialog>
+
+ColorLabel::ColorLabel(QWidget* parent, Qt::WindowFlags f)
+    : QLabel(parent, f)
+{
+    setFrameStyle(QFrame::Panel | QFrame::Sunken);
+    setFixedWidth(100);
+    setToolTip(tr("Click to change color."));
+}
+
+ColorLabel::~ColorLabel() {}
+
+void ColorLabel::setColor(const QColor& color)
+{
+    if (!color.isValid())
+        return;
+    stylesheetColor_ = color;
+    // ignore translucency
+    stylesheetColor_.setAlpha(255);
+    setStyleSheet(QStringLiteral("QLabel{background-color: rgb(%1, %2, %3);}")
+                  .arg(color.red()).arg(color.green()).arg(color.blue()));
+}
+
+QColor ColorLabel::getColor() const
+{
+    if (stylesheetColor_.isValid())
+        return stylesheetColor_; // the window color may be different from the stylesheet color
+    return palette().color(QPalette::Window);
+}
+
+void ColorLabel::mousePressEvent(QMouseEvent* /*event*/) {
+    QColor prevColor = getColor();
+    QColor color = QColorDialog::getColor(prevColor, window(), tr("Select Color"));
+    if (color.isValid() && color != prevColor)
+    {
+        emit colorChanged();
+        setColor(color);
+    }
+}
--- a/lxqt-config-appearance/colorLabel.cpp
+++ b/lxqt-config-appearance/colorLabel.cpp
@@ -45,7 +45,7 @@ void ColorLabel::setColor(const QColor& color)
     stylesheetColor_.setAlpha(255);
     QString borderColor = qGray(stylesheetColor_.rgb()) < 255 / 2
                             ? QStringLiteral("white") : QStringLiteral("black");
-    setStyleSheet(QStringLiteral("QLabel{background-color: rgb(%1, %2, %3); border: 1px solid %4;}}")
+    setStyleSheet(QStringLiteral("QLabel{background-color: rgb(%1, %2, %3); border: 1px solid %4;}")
                   .arg(color.red()).arg(color.green()).arg(color.blue()).arg(borderColor));
 }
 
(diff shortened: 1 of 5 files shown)


stefonarch:
Maybe there should be a neutral (white, grey) border after the  "window color:" view.
schermata-01-14-45-51

Ok, one sees the window color, but it looks empty.

tsujan:
> I see now that this makes it hard to see where to click to change again and is true only for black.

Yes. If the selected color is near the window color, the borders of the label might not be seen clearly. Any fixed color for the border would have the same problem with a range of selected colors.

I knew how to fix it but the fix wasn't trivial. Now that you mentioned it, I'll add the fix.

tsujan:
@stefonarch 
Added a distinguishable border to the color button. Please test again (with https://github.com/lxqt/lxqt-qtplugin/pull/58 and https://github.com/lxqt/libfm-qt/pull/545 if you don't want encounter any bug).

stefonarch:
Didn't look for bugs, but the border is ok now:
schermata-01-22-23-53
schermata-01-22-23-27

tsujan:
> Didn't look for bugs

Do look for them, please, but only with Fusion.

stefonarch:
After clicking "apply" the color below the selected item is blue like in other applications
schermata-02-14-27-27
And closing and reoping it   turns back like this:
schermata-02-14-27-14

-----------------------------
When changing from black to violet or other color some fields remain black always (non tested closing session), open file dialog, pcmanfm-qt (also after restarting desktop), config windows ecc
schermata-02-17-06-38
schermata-02-17-03-02

tsujan:
> After clicking "apply" the color below the selected item is blue like in other applications

That's a "problem" in Qt: The default palette Qt gives isn't that its default theme (Fusion) — in this case, "highlight" colors are different. Later we might try to find a workaround but, for now, I guess we have to tolerate it.

> When changing from black to violet or other color some fields remain black always (non tested closing session), open file dialog, pcmanfm-qt...

I didn't understand this completely. You might mean (1) the problem addressed at https://github.com/lxqt/libfm-qt/pull/545 (lack of contrast) or (2) the black base color (used by main views) or (3) the black disabled text color. The second and third problems are like what I mentioned above regarding the highlight color: they're what Qt gives to us as default colors, although they may be different from colors used by Qt's default theme (Fusion).

Anyway, these issues disappear once the app is restarted. Users might not like what I'm going to say but it's true: on-the-fly palette/style change is never perfect. A small part of the reason is explained in [Qt doc](https://doc.qt.io/qt-5/qapplication.html#setPalette):

"Some styles do not use the palette for all drawing, for instance, if they make use of native theme engines. This is the case for the Windows Vista and macOS styles."

It's the case for Kvantum too.

stefonarch:
> Anyway, these issues disappear once the app is restarted.  

No, I restarted now even the session. Setting to black, apply, setting ti violet, apply and restart session. For example the address bar in Browser and the areas windows as in the last 2 screens above remain black.

tsujan:
> For example the address bar in Browser and the areas windows as in the last 2 screens above remain black.

Then it's the color Qt automatically calculates when that window color is given to it. What's the problem? In your last 2 screenshots, all texts are clearly visible, except for the disabled ones, which aren't important.

tsujan:
→ Qt doc: https://doc.qt.io/qt-5/qpalette.html#QPalette-1 : "The other colors are..."

stefonarch:
> What's the problem? 

Let's say I want blue windows. 
White > blue: all the window is blue
Starting from black windows 
black > blue: windows (as posted) are half black still and will stay so always, unless I go back to white.

It's not about readability.

tsujan:
> black > blue: windows (as posted) are half black still and will stay so always, unless I go back to white.

It's related to the lightness/darkness of the window color. A black base color is what Qt adds; we only set the window color.

As was mentioned [elsewhere](https://github.com/lxqt/lxqt-qtplugin/blob/fc0db6dbd0cbe12421de93da23b6f5595f35438d/src/lxqtplatformtheme.cpp#L126), setting the window color is the minimal way of changing palettes. I might add more colors (to lxqt-qtplugin) later but the number should be kept small. 3 colors will be enough: Window, base and, probably, highlight colors.

Let's focus on setting only the window color for now. So far, so good.

tsujan:
On important thing: A light base color beside a dark window color would cause various problems. Therefore, even if the base color is made optional, the user should choose dark base colors (not necessarily black) with dark window colors. In short, he/she will be responsible for the problems he might cause with his choice.

lxqt-config-appearance/colorLabel.cpp
```diff
@@ -0,0 +1,67 @@
...
+void ColorLabel::setColor(const QColor& color)
+{
+    if (!color.isValid())
+        return;
+    stylesheetColor_ = color;
+    // ignore translucency
+    stylesheetColor_.setAlpha(255);
+    QString borderColor = qGray(stylesheetColor_.rgb()) < 255 / 2
+                            ? QStringLiteral("white") : QStringLiteral("black");
+    setStyleSheet(QStringLiteral("QLabel{background-color: rgb(%1, %2, %3); border: 1px solid %4;}}")
```
yan12125, line 48: Looks like the style sheet contains an extra curly bracket. Whenever the window color changes, a warning is printed in terminal:
```
Could not parse stylesheet of object ColorLabel(0x5580ec9e4200, name = "winColorLabel") 
```

yan12125:
> And closing and reoping it turns back like this:

I observed something different here (maybe because I'm on Qt 5.15?). After closing and reopening applications, highlighted items still uses darker blue

screen

By the way, Qt 5.15 has been moved to stable repositories on Arch.

lxqt-config-appearance/colorLabel.cpp
tsujan: Fixed now. Thanks!

Strangely, Qt 5.14 sets the stylesheet.

tsujan:
> After closing and reopening applications, highlighted items still uses darker blue

Yes, I [explained](https://github.com/lxqt/lxqt-config/pull/629#issuecomment-637630974) the reason to @stefonarch and also here: https://github.com/lxqt/lxqt-qtplugin/pull/56#issuecomment-636865071. In short, the default highlight color Qt calculates isn't the highlight color of its default style (= Fusion). That might be fixed when we add the highlight color to the customizable colors later.

yan12125:
Is it possible to add texts "Change..." to the ColorLabel? All other parts look good to merge ;)

tsujan:
> Is it possible to add texts "Change..." to the ColorLabel?

Yes, it is, with a small change to the code. I added a tooltip instead because it looked cleaner to me.

tsujan:
BTW, the highlight color Qt calculates is exactly `Qt::darkBlue` (= `#000080`), while Fusion's highlight color is `QColor(60, 140, 230)`. I think the second is more elegant. Will consider it as the default highlight color when adding highlight and base colors to lxqt-qtplugin; then, two other rows could be added to the Appearance dialog of lxqt-config in a future patch.

Apparently (→ [your comment](https://github.com/lxqt/lxqt-config/pull/629#issuecomment-638764333)), that will need Qt 5.15 to work correctly. I think Qt 5.15 will come to most non-rolling distros before LXQt 0.16 :)

luis-pereira:
@tsujan should this be tested in conjunction with https://github.com/lxqt/lxqt-qtplugin/pull/58 ?
I'm still on 5.14 tough.

tsujan:
@luis-pereira 
Preferably, yes, because, without that fix, you might see weird things in some apps (on Desktop, for example).

luis-pereira:
Before updating the apps (close/open):
p1

After updating:

p2

Is the difference in QTerminal expected ? 
p.s. sorry for the eye blinding colors.... but shades of gray are not easy to spot.

tsujan:
@luis-pereira That should be small bug in Qterminal, like the small bug in libfm-qt, which I fixed here: https://github.com/lxqt/libfm-qt/pull/545. In short, if an app changes the palette of one of its widgets, the app itself is responsible for updating that palette with a style change. Most apps (developers) don't care, though ;)

As I explained elsewhere, we should not force any palette from outside the app (see https://github.com/lxqt/lxqt-qtplugin/pull/58#issuecomment-638784055).

lxqt-config-appearance/colorLabel.cpp
```diff
@@ -0,0 +1,67 @@
...
+{
+    if (!color.isValid())
+        return;
+    stylesheetColor_ = color;
+    // ignore translucency
+    stylesheetColor_.setAlpha(255);
+    QString borderColor = qGray(stylesheetColor_.rgb()) < 255 / 2
+                            ? QStringLiteral("white") : QStringLiteral("black");
+    setStyleSheet(QStringLiteral("QLabel{background-color: rgb(%1, %2, %3); border: 1px solid %4;}")
+                  .arg(color.red()).arg(color.green()).arg(color.blue()).arg(borderColor));
```
luis-pereira, line 49: Minor improvement:
use multiple args instead of chained arg.

luis-pereira:
>  Most apps (developers) don't care, though ;)

Or think it's Qt job to handle that.

lxqt-config-appearance/colorLabel.cpp
tsujan: Is that possible? Three values are numbers and the fourth one is a string.

tsujan:
> Or think it's Qt job to handle that.

I'm afraid not. It's really the app's responsibility. Once a palette is explicitly set for a widget, that widget will become "isolated" from the app palette, so that, if the app palette is changed, the changes won't affect the brushes that are explicitly set for the widget in question.

This property of Qt is advantageous in various ways; for example, we use it for setting desktop text color.

lxqt-config-appearance/colorLabel.cpp
luis-pereira: For the first three, one can use `QString::number()`.

tsujan: Yes but wouldn't that be three extra computations that would defeat the purpose? Just asking because I'm not sure.

luis-pereira: Converting from number to string is always done. `QString::args(2)` will at some point fallback to `QString::number(2)`. 
When using chained args, there is one more allocation for each chained arg(). Using multiple args only one allocation is actually needed.

tsujan: > Converting from number to string is always done...

Reasonable.

Did as you told.

tsujan:
@yan12125, @luis-pereira 
I got Qt 5.15.0-2 today. After recompiling libqtxdg, liblxqt and lxqt-qtplugin against it (with all patches, this one included), I saw that Fusion's dark-blue problem is fixed. Here, the default highlight color is that of Fusion, namely `#3c8ce6`.

tsujan:
My mistake! The dark-blue problem is not fixed in Qt 5.15 but only reversed: while, in Qt 5.14, an on-the-fly color change resulted in a dark blue highlight color that was changed to `#3c8ce6` after app restart, in Qt 5.15, an on-the-fly color change creates `#3c8ce6` that is changed to dark-blue after app restart!

I hope the problem will be fixed in a patch version of Qt 5.15; it may show a deeper problem in Qt.

tsujan:
BTW, Qt 5.15. seems to be the cleanest Qt5 version.

Merging....

tsujan:
> BTW, Qt 5.15. seems to be the cleanest Qt5 version.

Replying to myself:

I should have waited a day or so before getting happy about Qt 5.15 (→ https://github.com/tsujan/Kvantum/issues/537). Since Qt 5.15 is supposed to be the last version of Qt5, I can say with certainty that there has been at least a bad regression (usually more) for each major release of Qt5.

PR #631, opened 2020-06-08
lxqt-config-appearance: Fix widget style chooser alignment

luis-pereira:
Make the label and combobox, position aligment and size behave.
Simply put them inside a layout.


luis-pereira:
Before:
p3
p4


After:
p5
p6

tsujan:
Yes, they seem better to me too in this way. GTM.

PR #634, opened 2020-06-18
Fix a crash due to 'output->currentMode()' being null sometimes

Nibelheims:
Hello,
I happen to happily use LXQT for a while now. However attempting to save my monitor settings always results in a crash of lxqt-config-monitor. I have to restore things manually every time I boot my computer, which is pretty annoying. FYI I'm using an external HDMI monitor connected to a laptop.
I have forked the code, rebuilt it in debug mode, and backtraced the crash.
Hope this can help a bit.
Best regards.

### DIFF
--- a/lxqt-config-monitor/kscreenutils.cpp
+++ b/lxqt-config-monitor/kscreenutils.cpp
@@ -17,11 +17,14 @@ void KScreenUtils::updateScreenSize(KScreen::ConfigPtr &config)
         if( !output->isConnected() )
             continue;
         QPoint pos = output->pos();
-        KScreen::ModePtr mode(output->currentMode());
-        width = qMax(pos.x() + mode->size().width(), width);
-        height = qMax(pos.y() + mode->size().height(), height);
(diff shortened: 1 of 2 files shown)


tsujan:
Thank you very much! I also encountered several crashes on saving, when I was experimenting with 2 monitors for testing other codes belonging to LXQt, but never found the time to investigate the crashes.

Your patch looks logical and good. I'll try it to see if fixes the crashes I encountered.

Nibelheims:
Hi,

Thank you for your reply. I’m glad to help.

By the way thank you for your work and for bringing a nice environment to
users. I use LXQt with KWin on Lubuntu 20.04 and the whole thing is a
delight.

I have to say that I didn’t dig into the whole code to check whether there
were some other uses of that risky pointer. Since I do not know the code
base I also cannot be sure if the patch is consistent with the rest. I did
check however that it solved my issues, of course.

Best regards

Nibelheims:
Thank you for your approval.

Best regards

PR #635, opened 2020-07-08
Fixed type associations and added "Default Application"

tsujan:
This is a relatively big code change:

 1. Scheme protocols are added to the end of associations list.
 2. A tab for default apps is added. For now, it supports the default browser and email client.
 3. Default associations are done in a DE-specific way, so that they don't affect other DEs.
 4. The current changes can be undone correctly and reliably.
 5. The deprecated class `XdgDesktopFileCache` and the inefficient class `LXQt::SettingsCache` are removed from the code. `XdgMimeApps` is used instead of the former. The latter had bugs and couldn't do a correct and clean resetting; `QTemporaryFile` is used instead of it.
 6. All memory leaks are fixed. They were found both by rereading the code and by using `valgrind`.

NOTE 1: The patch relies on the latest git `libqtxdg` and SHOULD NOT be used without it.

NOTE 2: `lxqt-config-file-associations` only supports setting of default apps. The support for removing/sorting associations is still missing. Association removal is supported by `libqtxdg` → `XdgMimeApps` and sorting can be added to it.

Closes https://github.com/lxqt/lxqt/issues/1513

### DIFF
--- a/lxqt-config-file-associations/applicationchooser.cpp
+++ b/lxqt-config-file-associations/applicationchooser.cpp
@@ -32,6 +32,7 @@
 #include <QTimer>
 
 #include <XdgDesktopFile>
+#include <XdgMimeApps>
 
 #include <algorithm>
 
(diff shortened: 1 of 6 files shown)


tsujan:
So fast!

> The Default Apps get and set check code (the browser section) led me to think, that that code and criteria should be in a lib.

I can't think of another use for it but if you believe that it should be in the lib, it's OK with me.

IMHO, `LXQt::SettingsCache` needs more attention (→ https://github.com/lxqt/liblxqt/issues/262). I doubt that it can be fixed based on `QSettings` because its problem is inherited from `QSettings`. Maybe, it can be replaced by `QTemporaryFile` everywhere, in which case, it might not be needed at all.

Issue #402, opened 2018-08-24
[lxqt-config] Add HiDPI configuration


tsimonq2:
Carrying over from #916, we need a menu for HiDPI configuration.

tsujan:
> We should really work on this - not only talk.

A little more talk and I'll shut my mouth: LXQt has a nice GUI for settings EVs. There are 2 HDPI variables, one for Qt and the other for GTK+. Not-so-advanced users are aware of them and could set them. So, LXQt has no practical problem with HDPI.

A separate GUI may be good for newbies; nothing more.

And please don't mistake HDPI setting with font/icon size setting!

shlyakpavel:
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/19782228/63799940-c817a600-c915-11e9-8b68-27d189e7fdfd.png">

That's the default LXQt on my 13 inch 261ppi screen :(

agaida:
imagine you had a really high resolution screen :D

tsujan:
@shlyakpavel Qt has something called "scale factor". Consult Qt doc.

tsujan:
Example:

![scale_factor](https://user-images.githubusercontent.com/981076/63805798-74b25180-c92f-11e9-8bd5-a7d243f31a82.png)

renom:
It'd be great to be able to change DPI settings without restarting everything. It can already be done for GTK apps in GNOME/Cinnamon. Is it possible to have this for Qt and GTK apps in Lxqt?

Use case: a laptop with Full HD screen, an external monitor is 4K UHD. Since there's no way to have per-monitor settings without Wayland, the only solution I've found is to change DPI settings depending on the current screen.

tsujan:
> It'd be great to be able to change DPI settings without restarting everything.

It depends on whether Qt supports a runtime update (I don't think it does). If it does, lxqt-qtplugin might be able to support it.

For now, we need a GUI for users not to worry about the names of EVs.

renom:
@tsujan it's not related with changing DPI settings on the fly, but I've found out that there's a solution to have per-monitor scaling only for Qt apps with `QT_SCREEN_SCALE_FACTORS="DP-1=2;eDP-1=1.5"`. AFAIK, GTK doesn't have something similar to this. But since Lxqt is based on Qt, you could add a GUI to adjust this EV.

tsujan:
@renom Thanks for the info.

[...]

PR #636, opened 2020-07-20
Better look for color label in lxqt-config-appearance

tsujan:
The label is painted as the color label of `QColorDialog`, without using style sheets. This way, its look is more professional and consistent with the widget style.

### DIFF
--- a/lxqt-config-appearance/colorLabel.cpp
+++ b/lxqt-config-appearance/colorLabel.cpp
@@ -25,13 +25,17 @@
 
 #include "colorLabel.h"
 #include <QColorDialog>
+#include <QStyleOptionFrame>
+#include <QPainter>
 
 ColorLabel::ColorLabel(QWidget* parent, Qt::WindowFlags f)
     : QLabel(parent, f)
 {
     setFrameStyle(QFrame::Panel | QFrame::Sunken);
+    setLineWidth(1);
     setFixedWidth(100);
     setToolTip(tr("Click to change color."));
+    color_ = palette().color(QPalette::Window);
 }
 
 ColorLabel::~ColorLabel() {}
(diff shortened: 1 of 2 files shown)


lxqt-config-appearance/colorLabel.cpp
```diff
@@ -25,13 +25,17 @@
...
 #include "colorLabel.h"
 #include <QColorDialog>
+#include <QStyleOptionFrame>
+#include <QPainter>
 
 ColorLabel::ColorLabel(QWidget* parent, Qt::WindowFlags f)
     : QLabel(parent, f)
 {
     setFrameStyle(QFrame::Panel | QFrame::Sunken);
+    setLineWidth(1);
```
shlyakpavel, line 35: Why?

lxqt-config-appearance/colorLabel.cpp
```diff
@@ -25,13 +25,17 @@
...
     setFixedWidth(100);
```
shlyakpavel, line 36: Do we still have fixed sizes?

shlyakpavel: Is it okay with? https://github.com/lxqt/lxqt-config/issues/402
Scale factor seems to be no excuse to avoid default sizes

tsujan:
@shlyakpavel 
A border of 1px is actually 2px with a scale factor of 2. When you make a GUI for an app that has `Qt::AA_UseHighDpiPixmaps`, you shouldn't think about scale factors: X always means X multiplied by scale factor. All LXQt apps have `Qt::AA_UseHighDpiPixmaps`.

luis-pereira:
Didn't notice this. Will do it ASAP.

lxqt-config-appearance/colorLabel.cpp
luis-pereira: What would a different implementation looks like ?

tsujan: @luis-pereira 
If you asked me, I didn't understand your question.

IMHO, avoiding stylesheets and respecting the widget style is a good practice. For that, I just did what Qt does in its color dialog.

As for the scale factor, it doesn't have any relevance here.

luis-pereira: @tsujan I was asking @shlyakpavel how he would implement the `setLineWith`. The stylessheet stuff is self-evident.

shlyakpavel: Agree on stylesheet being wrong. However, would stick to the default border.

tsujan: @shlyakpavel 
The default border width *is* 1. Here, I was only explicit about it because I think it's good to be explicit.

shlyakpavel: > @shlyakpavel
> I think it's good to be explicit.

I don't. This may vary between systems so we should better keep it to look native and integrated.

tsujan: @shlyakpavel 
Qt doc: https://doc.qt.io/qt-5/qframe.html#lineWidth-prop. Quoting: "The default value is 1." And it has nothing to do with integration or native look. *This PR* does.

A visual comparison that shows the border is identical to that of Qt's color dialog:

border

shlyakpavel: @tsujan ok, it's up to you

PR #639, opened 2020-08-03
lxqt-config-brightness: Initialize members in the constructor

luis-pereira:
Every member should be initialize. Prevents bugs and helps to find others.

### DIFF
--- a/lxqt-config-brightness/monitorinfo.cpp
+++ b/lxqt-config-brightness/monitorinfo.cpp
@@ -19,10 +19,12 @@
 #include "monitorinfo.h"
 
 MonitorInfo::MonitorInfo(int id, QString name, long backlightMax)
+    : mBacklightMax(backlightMax),
+      mBacklight(-1),
+      mBrightness(-1.0f),
+      mName(name),
+      mId(id)
 {
-    mId = id;
-    mName = name;
-    mBacklightMax = backlightMax;
 }
 
 MonitorInfo::MonitorInfo(const MonitorInfo &monitor)


lxqt-config-brightness/monitorinfo.cpp
```diff
@@ -19,10 +19,12 @@
 #include "monitorinfo.h"
 
 MonitorInfo::MonitorInfo(int id, QString name, long backlightMax)
+    : mBacklightMax(backlightMax),
+      mBacklight(-1),
+      mBrightness(-1.0f),
```
tsujan, line 24: Shouldn't this be `1.0f`? `mBacklight` can be negative ("negative" means "not supported") but `mBrightness` is between `0` and `2`.

luis-pereira: My rationale was to initialize `mBrightness` with an impossible value to distinguish it from the ones set with `setBrightness()`

tsujan: I didn't mean that it's a real problem; the PR is good. IMO, "1.0f" for brightness (=100% brightness) is just a little more logical as the default.

Anyway: GTM.

luis-pereira: > I didn't mean that it's a real problem; the PR is good. IMO, "1.0f" for brightness (=100% brightness) is just a little more logical as the default.
> 
> Anyway: GTM.

I do understand your point. Since the correct/current value is unknown setting it to a possible value seems odd to me.
Maybe the range is 0.0f to 2.0f to accommodate gamma when backlight is not available.

luis-pereira:
Changed commit message. Replaced lxqt-config-appearance with lxqt-config-brightness.

PR #641, opened 2020-08-31
Use XdgDefaultApps to set default applications

tsujan:
Also fixed a problem in setting the default browser and added a section for the default file manager (because @luis-pereira had added it to `XdgDefaultApps`).

### DIFF
--- a/lxqt-config-file-associations/applicationchooser.cpp
+++ b/lxqt-config-file-associations/applicationchooser.cpp
@@ -33,6 +33,7 @@
 
 #include <XdgDesktopFile>
 #include <XdgMimeApps>
+#include <XdgDefaultApps>
 
 #include <algorithm>
 
@@ -40,10 +41,29 @@
 
 Q_DECLARE_METATYPE(XdgDesktopFile*)
 
-ApplicationChooser::ApplicationChooser(const QString& type, bool showUseAlwaysCheckBox)
+ApplicationChooser::ApplicationChooser(const QString& type, int cat)
 {
     widget.setupUi(this);
-    m_Type = type;
+
+    m_Cat = cat;
+    switch (cat)
+    {
+        case category::webBrowser :
+            m_Type = QStringLiteral("x-scheme-handler/http");
+            break;
+        case category::emailClient :
+            m_Type = QStringLiteral("x-scheme-handler/mailto");
+            break;
+        case category::fileManager :
+            m_Type = QStringLiteral("inode/directory");
+            break;
+        default :
+            m_Cat = category::none;
+            break;
+    }
+
+    if (m_Type.isEmpty())
+        m_Type = type;
 
     XdgMimeApps appsDb;
     m_CurrentDefaultApplication = appsDb.defaultApp(m_Type);
@@ -60,7 +80,6 @@ ApplicationChooser::ApplicationChooser(const QString& type, bool showUseAlwaysCh
         widget.mimetypeLabel->hide();
     }
 
-    widget.alwaysUseCheckBox->setVisible(showUseAlwaysCheckBox);
     widget.buttonBox->button(QDialogButtonBox::Ok)->setEnabled(false);
 }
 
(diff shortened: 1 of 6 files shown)


luis-pereira:
@tsujan Wasn't aware. Will review it.

luis-pereira (commented):
- The flle manager part should not probably be included, yet, since it will create a bad user experience, e.g. pcmanfm-qt doesn't considers it. 

- It's good practice to not use the 3 argument connect() that takes a lambda. We should use the overload with 4 arguments, taking a context object to control the connection’s lifetime. Clazy warns about it and it has a point. A simple "fix":
 ```diff
 diff --git a/lxqt-config-file-associations/mimetypeviewer.cpp b/lxqt-config-file-associations/mimetypeviewer.cpp
 index dd4da3e..ebefb49 100644
 --- a/lxqt-config-file-associations/mimetypeviewer.cpp
 +++ b/lxqt-config-file-associations/mimetypeviewer.cpp
 @@ -179,7 +179,7 @@ MimetypeViewer::MimetypeViewer(QWidget *parent)
  
      // "Default Applications" tab
      updateDefaultApplications();
 -    connect(widget.chooseBrowserButton, &QAbstractButton::clicked, [this]() {
 +    connect(widget.chooseBrowserButton, &QAbstractButton::clicked, this, [this]() {
          if (XdgDesktopFile *app = chooseApp(QStringLiteral("x-scheme-handler/http")))
          {
              XdgDesktopFile *defaultBrowser = XdgDefaultApps::webBrowser();
 @@ -192,7 +192,7 @@ MimetypeViewer::MimetypeViewer(QWidget *parent)
              delete app;
          }
      });
 -    connect(widget.chooseEmailClientButton, &QAbstractButton::clicked, [this]() {
 +    connect(widget.chooseEmailClientButton, &QAbstractButton::clicked, this, [this]() {
          if (XdgDesktopFile *app = chooseApp(QStringLiteral("x-scheme-handler/mailto")))
          {
              XdgDesktopFile *defaultEmailClient = XdgDefaultApps::emailClient();
 @@ -205,7 +205,7 @@ MimetypeViewer::MimetypeViewer(QWidget *parent)
              delete app;
          }
      });
 -    connect(widget.chooseFileManagerButton, &QAbstractButton::clicked, [this]() {
 +    connect(widget.chooseFileManagerButton, &QAbstractButton::clicked, this, [this]() {
          if (XdgDesktopFile *app = chooseApp(QStringLiteral("inode/directory")))
          {
              XdgDesktopFile *defaultFileManager = XdgDefaultApps::fileManager();
 
 ```
 
- `XdgDefaultApps` should provide a `QList<XdgDesktopFile *> XdgDefaultApps::webBrowsers()` providing all the web browsers installed. The same for the other categories. I'm already testing it. It avoids issues like Qmmp (a media player) being presented as a file manager.
 screenshot-1

- A distinct special application chooser improves the user experience, IMO. That said, we are using what we have now. It can be improved later on.

tsujan:
@luis-pereira 
**When there is no context**, the context argument in lambdas is not only good but necessary. Here, `widget.chooseBrowserButton` (and the 2 other widgets) serve as contexts because they are deleted before `this` -- there's no need to an extra `this`.

As for the file manager, I was against its inclusion at first (don't remember the page, where we discussed it) but, after you added it, I think it's needed when opening links inside apps that support local links (like FeatherNotes or any other notes manager).

Neither pcmanfm-qt nor any other FM should consider the default file manager because it's itself a FM. It's like opening a PNG file from inside an image viewer with another app that opens PNG files by default.

I agree that *desktop* could create a little confusion because many users may not know that it belongs to pcmanfm-qt. But that shouldn't prevent us from defining a default file manager -- the "desktop issue" could be fixed separately, if it's really an issue.

Finally, if Qnmp adds `inode/directory` to its desktop file, it'll create a problem in *all* DEs and with *all* file managers. That should be fixed in Qnmp's desktop file and, IM**V**HO, we shouldn't do anything about it.

tsujan:
> Clazy warns about it and it has a point

Oh, I didn't see it. Never used Clazy but I think it needs improvement.

luis-pereira:
> **When there is no context**, the context argument in lambdas is not only good but necessary. Here, `widget.chooseBrowserButton` and the 2 other widgets serve as contexts because they are deleted before `this` -- there's no need to an extra `this`.

@tsujan I know all of that. But clazy still complains about it. Signficant parts of KDE did this change because most probably clazy will not fix it in the near future. Adding the extra 'this`doesn't have any side-effects. That was just a suggestion.

 
> As for the file manager, I was against its inclusion at first (don't remember the page, where we discussed it) but, after you added it, I think it's needed when opening links inside apps that support local links (like FeatherNotes or any other notes manager).
> 
> Neither pcmanfm-qt nor any other FM should consider the default file manager because it's itself a FM. It's like opening a PNG file from inside an image viewer with another app that opens PNG files by default.

A file manager opens dirs on it's own. There is no discussion about that.
 
> I agree that _desktop_ could create a little confusion because many users may not know that it belongs to pcmanfm-qt. But that shouldn't prevent us from defining a default file manager -- the "desktop issue" could be fixed separately, if it's really an issue.

The desktop issue misleads the user. The fact that the desktop is part of a file manager is a implementation detail.
The default file manager should be include after the "desktop issue" has been solved, IMO.
 
> Finally, if Qnmp adds `inode/directory` to its desktop file, it'll create a problem in _all_ DEs and with _all_ file managers. That should be fixed in Qnmp's desktop file and, IM**V**HO, we shouldn't do anything about it.

Qmmp handles directories. You can open a dir with qmmp, so they are not wrong. But qmmp is not a file manager.
We have the same issue with the web browser:
screen1

SpaceFM is not a web browser. Supporting `x-scheme-hander/http` is not the definition of a web browser. We should use the same criteria as to determine if an app is the default browser. The use of Categories provides a better user experience.

tsujan:
@luis-pereira 
Problems like that of Qmmp will be solved [by your PR](https://github.com/lxqt/libqtxdg/pull/228) and a modification here. As I said elsewhere, considering categories was a good idea of yours that I'd failed to see.

> But clazy still complains about it. Signficant parts of KDE did this change because most probably clazy will not fix it

A tool that warns about contexts should be smart enough to see them. Apparently, Clazy isn't so.

> The default file manager should be include after the "desktop issue" has been solved, IMO.

I think it's too important to be ignored. Any app may need it; for example, a browser may need to open the download folder, or I added an item to FeatherPad's tab context menu for opening the containing folder, etc.

luis-pereira:
> @luis-pereira
> Problems like that of Qmmp will be solved [by your PR](https://github.com/lxqt/libqtxdg/pull/228) and a modification here. As I said elsewhere, considering categories was a good idea of yours that I'd failed to see.

I wrote this comment without reading your comment there.

> I think it's too important to be ignored. Any app may need it; for example, a browser may need to open the download folder, or I added an item to FeatherPad's tab context menu for opening the containing folder, etc.

It's important. But the ironic thing about it, is that the Desktop will be the one to not use the default manager.

tsujan:
> It's important. But the ironic thing about it, is that the Desktop will be the one to not use the default manager.

I know, have opened https://github.com/lxqt/lxqt/issues/1823 for it, and will try to deal with it in pcmanfm-qt's code later. But, because of some (new) options in pcmanfm-qt, opening desktop folders with another file manager isn't as trivial as it may seem. So, I don't think it'll be ready soon, while the default file manager is too important in other places.

tsujan:
The desktop issue aside, one practical thing about the default file manager is that most LXQt users want to open folders from other apps *in pcmanfm-qt* but they may have installed other file managers too.

tsujan:
@luis-pereira 
I added a commit to use your [latest libqtxdg commit](https://github.com/lxqt/libqtxdg/commit/0133926cfed4f2f1c1e407b3592786db6cf85ed1). Please test it whenever you find the time.

PR #645, opened 2020-09-15
add xinput driver detection for touchpad devices

### LINKED ISSUE #644: All Touchpad settings disabled
##### Expected Behavior

Ability to adjust touchpad properties.

##### Current Behavior

One Picture, less words. :)

touchpad-config-disabled

##### Possible Solution
Show warning when driver != "libinput".

##### Steps to Reproduce (for bugs)
1. Open lxqt-config-input
2. select touchpad settings

##### Context

Trying to customize Touchpad to my needs.

##### System Information
* Distribution & Version: Manjaro 20.1
* Kernel: 5.4 (LTS)
* Qt Version: 5.15
* liblxqt Version: 0.15.1
* lxqt-build-tools Version: ?? 
* Package version: 0.15.0

antis81:
As the driver name cannot be queried directly from xinput api we use a "known" property instead.

Here's an example how it looks like:

lxqt-config-input

The info is translatable and only visible, if `xinput driver != libinput`.

Closes #644

### DIFF
--- a/lxqt-config-input/touchpadconfig.cpp
+++ b/lxqt-config-input/touchpadconfig.cpp
@@ -86,8 +86,8 @@ void TouchpadConfig::initControls()
               "LXQt only supports \"libinput\" as xinput driver.\n"
               "(current value: %1)\n"
               "\n"
-              "Check your xinput configuration (/etc/X11/xorg.conf.d).\n"
-              "If not needed, also remove the xf86-input-synaptics package.")
+              "If this is intended, please configure xinput manually.\n"
+              "Otherwise you can get rid of this message by changing xinput driver to \"libinput\".\n")
             .arg(device.xinputDriver()));
       ui.deviceInfoLabel->setStyleSheet(QStringLiteral("background-color: #f88; border-color: #f33; border-size: 1px"));
     };
--- a/lxqt-config-input/touchpadconfig.cpp
+++ b/lxqt-config-input/touchpadconfig.cpp
@@ -79,6 +79,21 @@ void TouchpadConfig::initControls()
     initFeatureControl(ui.naturalScrollingEnabledCheckBox, device.naturalScrollingEnabled());
     initFeatureControl(ui.tapToDragEnabledCheckBox, device.tapToDragEnabled());
 
+    auto ok = device.xinputDriverSupported();
+    if (!ok) {
+      ui.deviceInfoLabel->setText(
+            tr(
+              "LXQT only supports \"libinput\" as xinput driver.\n"
+              "(current value: %1)\n"
+              "\n"
+              "Check your xinput device configuration (/etc/X11/xorg.conf.d).\n"
+              "If not needed, also remove the xf86-input-synatics package.")
+            .arg(device.xinputDriver()));
+      ui.deviceInfoLabel->setStyleSheet(QStringLiteral("background-color: #f88; border-color: #f33; border-size: 1px"));
+    };
+    ui.deviceInfoLabel->setVisible(!ok);
+    ui.deviceInfoLabel->setFrameStyle(QFrame::Panel | QFrame::Sunken);
+
     float accelSpeed = device.accelSpeed();
     if (!std::isnan(accelSpeed)) {
         ui.accelSpeedDoubleSpinBox->setEnabled(true);
--- a/lxqt-config-input/touchpadconfig.cpp
+++ b/lxqt-config-input/touchpadconfig.cpp
@@ -83,11 +83,11 @@ void TouchpadConfig::initControls()
     if (!ok) {
       ui.deviceInfoLabel->setText(
             tr(
-              "LXQT only supports \"libinput\" as xinput driver.\n"
+              "LXQt only supports \"libinput\" as xinput driver.\n"
               "(current value: %1)\n"
               "\n"
-              "Check your xinput device configuration (/etc/X11/xorg.conf.d).\n"
-              "If not needed, also remove the xf86-input-synatics package.")
+              "Check your xinput configuration (/etc/X11/xorg.conf.d).\n"
+              "If not needed, also remove the xf86-input-synaptics package.")
             .arg(device.xinputDriver()));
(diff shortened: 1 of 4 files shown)


yan12125:
Thanks for this patch! Here are some comments:

* It is better for `m_xinputDriver` to use libinput/non-libinput instead of libinput/synaptics. Maybe there are other drivers.
* LXQT -> LXQt
* I'm not sure if the second paragraph applies to all scenarios. I remember if only the synaptics driver is installed, Xorg automatically picks it, so users might get confused about `/etc/X11/xorg.conf.d`. Also that path might be different on another system.
* Other systems may use a different package name than `xf86-input-synaptics`.

antis81:
>It is better for m_xinputDriver to use libinput/non-libinput instead of libinput/synaptics. Maybe there are other drivers.

Have a look at this:

```bash
~ >>> xinput --list                                                                                                                                                                                             
⎡ Virtual core pointer                          id=2    [master pointer  (3)]
⎜   ↳ SynPS/2 Synaptics TouchPad                id=12   [slave  pointer  (2)]
⎜   ↳ Logitech USB-PS/2 Optical Mouse           id=13   [slave  pointer  (2)]
…

~ >>> xinput --list-props 12                                                                                                                                                                                    
Device 'SynPS/2 Synaptics TouchPad':
        Device Enabled (170):   1
        Coordinate Transformation Matrix (172): 1.000000, 0.000000, 0.000000, 0.000000, 1.000000, 0.000000, 0.000000, 0.000000, 1.000000
        libinput Tapping Enabled (308): 1
        libinput Tapping Enabled Default (309): 0
        libinput Tapping Drag Enabled (310):    1
        libinput Tapping Drag Enabled Default (311):    1
        libinput Tapping Drag Lock Enabled (312):       0
        libinput Tapping Drag Lock Enabled Default (313):       0
        libinput Tapping Button Mapping Enabled (314):  1, 0
        libinput Tapping Button Mapping Default (315):  1, 0
        libinput Natural Scrolling Enabled (316):       1
        libinput Natural Scrolling Enabled Default (317):       0
        libinput Disable While Typing Enabled (318):    1
        libinput Disable While Typing Enabled Default (319):    1
        libinput Scroll Methods Available (320):        1, 1, 0
        libinput Scroll Method Enabled (321):   1, 0, 0
        libinput Scroll Method Enabled Default (322):   1, 0, 0
        libinput Accel Speed (323):     0.000000
        libinput Accel Speed Default (324):     0.000000
        libinput Accel Profiles Available (325):        1, 1
        libinput Accel Profile Enabled (326):   1, 0
        libinput Accel Profile Enabled Default (327):   1, 0
        libinput Left Handed Enabled (328):     0
        libinput Left Handed Enabled Default (329):     0
        libinput Send Events Modes Available (293):     1, 1
        libinput Send Events Mode Enabled (294):        0, 0
        libinput Send Events Mode Enabled Default (295):        0, 0
        Device Node (296):      "/dev/input/event15"
        Device Product ID (297):        2, 7
        libinput Drag Lock Buttons (330):       <no items>
        libinput Horizontal Scroll Enabled (331):       1

# TODO: synaptics will follow (requires reinstallation of deprecated package & reboot)
```

Do you see the problem with that idea? 

I haven't found a way to explicitly query the xinit plugin name (why?). So detecting the deprecated synaptics xinit plugin by checking on a "known attribute" is reliable. Detecting libinput however can produce a "false positive". This happens if both are installed (which is the common case, especially when upgrading LTS based distro's or rolling releases).

And we haven't even taken Wayland into account, where none of libinput/synaptics is being used and all this will fail… Probably wrapping [QTouchDevice](https://doc.qt.io/qt-5/qtouchdevice.html) would be a better way instead of trying to reinvent the wheel.

yan12125:
> I haven't found a way to directly query the xinit plugin name from xinput `Device Driver` key

Yep unfortunately that seems the case. A method is checking the existence of at least one of `LIBINPUT_PROP_*` property in `touchpaddevice.cpp`. That will be non-trivial for sure.

> And we haven't even taken Wayland into account

Actually we do ;) libinput was originally a sub-project of Wayland and is still used by many Wayland compositors. Hopefully adding an abstraction layer to switch between xinput/wayland can magically make this dialog work on Wayland.

antis81:
Is there anything holding you back from merging?

Concerning your questions about the hint-text:
It certainly helps "beginners" to fix their system and I like to keep it. The location and package name both are correct for a broad variety of today's distributions (at least Arch, Debian >=9, Ubuntu >=18.04 are covered). Especially on very young Distro's (e.g "Endeavour", which also advertises LXQt) the issue is practically non-existant and libinput (or Wayland) is being installed right from start. Which other distro's you have in mind? And are those really relevant and targeted by LXQt?

yan12125:
> at least Arch, Debian >=9, Ubuntu >=18.04 are covered

Maybe you got incorrect or out-dated information? From Repology [1], the package name is `xserver-xorg-input-synaptics` on all currently-supported versions of Debian and its derivatives. And for distributions from Red Hat (CentOS, Fedora), it is `xorg-x11-drv-synaptics`.

> Which other distro's you have in mind? And are those really relevant and targeted by LXQt?

LXQt aims to be distribution- and kernel-agnostic. I will not call a system "not relevant" with LXQt. Among active maintainers/participants of the LXQt project, there are at least Arch and derivatives, Gentoo, OpenSUSE, FreeBSD, NetBSD and Ubuntu users. (there are definitely more - I don't have that much memory :P) Adding names and paths for specific distributions does not sound a good idea.

By the way, FreeBSD appears to put X11 configuration files to /usr/local/etc/X11 [2].

[1] https://repology.org/project/xdrv:synaptics/versions
[2] https://www.freebsd.org/doc/handbook/x-config.html

antis81:
Thanks for pointing that out. I am going to change the hint text to something generic (or leave out completely) so this can be merged. (One of those things where Linux/*nix shows it's "ugly" - political - side. :8ball: :slightly_smiling_face:)

yan12125:
Thanks!

PR #649, opened 2020-10-08
Fix possible null pointer dereference

luis-pereira:
Either crtc_info == NULL is redundant or there is a possible null pointer
dereference.

Closes https://github.com/lxqt/lxqt/issues/1831.


tsujan (approved):
GTM

PR #654, opened 2020-10-18
Remember sizes of appearance and file associations dialogs

tsujan:
Other dialogs should also remember their sizes but the above ones need it more.


stefonarch:
Works ;)

tsujan:
@stefonarch, thanks for testing!

The code is trivial; merging...

Issue #676, opened 2020-11-09
lxqt-config-input (v0.16.0) > Mouse and Touchpad > Acceleration speed > Unsupported data type QVariant::float


marc-sabates-gelpi:
##### Expected Behavior

The acceleration speed should be set properly and kept to what the user is setting it to.

##### Current Behavior

When setting a value for the acceleration, (and when the program is started from the CLI), there is the message `Unsupported data type QVariant::float`, and the value is reset to zero.

##### Possible Solution

Perhaps there is a type conversion issue, I don't know.
Having said that, when the config is set manually with `xinput` it is displayed correctly on `lxqt-config-input`.

##### Steps to Reproduce (for bugs)

1. Open `lxqt-config-input` in the CLI.
2. Click on `Mouse and Touchpad`.
3. Set a value for _Acceleration speed_, e.g. `0.6`.
4. Press the `Apply` button.
5. `Unsupported data type QVariant::float` appears in the console (in the background), and the acceleration speed has been set to _zero_, yet the displayed acceleration is the correct one.

When closing and opening the config again the value displayed is `0.00`.

##### Context

After the last update (v0.16.0) the trackpad speed was terribly slow and trying to set it to a higher value wouldn't work.
I am setting it manually from the CLI with `xinput --set-prop 'SynPS/2 Synaptics TouchPad' 'libinput Accel Speed' 0.5`

##### System Information

* Distribution & Version: ArchLinux 
* Kernel: 5.9.6-arch1-1 #1 SMP PREEMPT x86_64 GNU/Linux 
* Qt Version: 0.16.0
* liblxqt Version: 0.16.0
* lxqt-build-tools Version: 0.8.0
* Package version: 0.16.0-1

The whole ouput on my console when running lxqt-config-input is:
```
$ lxqt-config-input
add "gb" ""
Detected "SynPS/2 Synaptics TouchPad" on "/dev/input/event11"
xinput driver: "libinput"
Detected "TPPS/2 IBM TrackPoint" on "/dev/input/event12"
xinput driver: "libinput"
Unsupported data type QVariant::float
```

Thanks guys!

tsujan:
Although I can't see "Unsupported data type QVariant::float", I confirm that the accelerations rests.

Thanks for the report!

tsujan:
It was a regression. Its fix is here https://github.com/lxqt/lxqt-config/pull/677. Most probably, lxqt-config 0.16.1 will be releases with it soon.

[...]

PR #677, opened 2020-11-10
Fixed regression in setting acceleration

tsujan:
The regression was introduced in https://github.com/lxqt/lxqt-config/pull/645, probably because `Float` wasn't a member of `QVariant`. (Dear contributors, please don't include unrelated changes into a single PR!)

The patch fixes the problem by using`QMetaType::Type`. As Qt doc says, the return value of `QVariant::type()` should be interpreted as `QMetaType::Type` (see https://doc.qt.io/qt-5/qvariant.html#type).

Fixes https://github.com/lxqt/lxqt-config/issues/676

@yan12125, we'll need a point release if you approve the PR.

### DIFF
--- a/lxqt-config-input/touchpaddevice.cpp
+++ b/lxqt-config-input/touchpaddevice.cpp
@@ -108,10 +108,10 @@ static bool xi2_set_device_property(int deviceid, const char* prop, QList<QVaria
 
     XFree(data);
 
-    auto dataType = values[0].type();
+    auto dataType = static_cast<QMetaType::Type>(values[0].type());
     switch (dataType)
     {
-    case QVariant::Int:
+    case QMetaType::Int:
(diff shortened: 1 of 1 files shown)


yan12125:
Thanks a lot for the fix and detailed investigation in Qt documents. I was confused by QVariant::Type and QMetaType, and thought they are exchangeable.

tsujan:
> I was confused by QVariant::Type

It *is* confusing. Qt is self-inconsistent in this regard. It shows something (`QVariant::Type`) but uses something else (`QMetaType::Type`). The lack of `QVariant::Float` is the worst part of it. Qt doc just hides the issue behind a technical, vague explanation.

yan12125:
Indeed. Sigh.

PR #681, opened 2020-11-19
translations: Use .yaml for .desktop translation

palinek:


stefonarch:
Checked all files now.

To do:
 convert in .yaml the  *directory files in src/menuname/translations

PR #692, opened 2020-12-15
Drop obsolete QProcess::start()

luis-pereira:


tsujan:
GTM

Haven't tried it but it should be OK — unless a magic happens.

luis-pereira:
Using it without issues.

PR #704, opened 2021-01-11
load default color palette from theme

antis81:
The "Default Palette" button now loads the color palette from the current theme (if it provides a `lxqt-palette.conf` file).
The previous default palette is used as a fallback.

### DIFF
--- a/lxqt-config-appearance/styleconfig.cpp
+++ b/lxqt-config-appearance/styleconfig.cpp
@@ -72,12 +72,11 @@ StyleConfig::StyleConfig(LXQt::Settings* settings, QSettings* qtSettings, LXQt::
     connect(ui->linkVisitedColorLabel, &ColorLabel::colorChanged, this, &StyleConfig::settingsChanged);
 
     connect(ui->defaultPaletteBtn, &QAbstractButton::clicked, [this] {
-        QColor winColor(239, 239, 239);
-        QPalette defaultPalette(winColor);
-        ui->winColorLabel->setColor(winColor, true);
+        //auto defaultPalette = this->defaultPalette();
+        auto defaultPalette = loadPalette(QStringLiteral("%1/lxqt-palette.conf").arg(LXQt::LXQtTheme::currentTheme().path()));
+        ui->winColorLabel->setColor(defaultPalette.color(QPalette::Window), true);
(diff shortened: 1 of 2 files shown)


tsujan (changes requested):
This isn't what we want. The default palette shouldn't depend on anything. People can use `lxqt-config` without using lxqt-panel or anything that would need an LXQt theme.

What we want is just applying a palette when an LXQt theme is applied in the "LXQt Theme" page. The "Widget Style" page should be updated to show it but the default palette isn't related to this.

antis81:
Yeah I know! :slightly_smiling_face:

tsujan:
There are at least two ways of doing it:

 1. The hard way is adding a check box to the "LXQt Theme" page that says, "Also, apply Qt palettes of themes". If it's checked (it should be unchecked by default), when the user applies an LXQt theme, its corresponding Qt palette will be applied (if existing) and the "Widget Style" page will reflect that.
 2. The easier way is adding a button below "Default Palette" in the "LXQt Theme" page. The text of that button can be, "LXQt Theme Palette". If it's clicked, the Qt Palette corresponding to the active LXQt theme will be applied (if existing) .

The second way needs less code and does the job but it needs the user to apply the palette — which seems acceptable to me.

tsujan:
I really recommend the second way mentioned above — it's simple and respects the user's choice — unless you have a better way in mind.

Another reason why I prefer the second method is that, unlike KDE, LXQt doesn't have a default Qt style to worry about automatic applying of palettes, to say nothing of the fact that some Qt styles (like Kvantum and gtk2) ignore external palettes.

tsujan:
EDIT: I corrected the description of the second method: I meant the "Widget Style" page but wrote "LXQt Theme".

antis81:
> …unless you have a better way in mind.

Guess what, I do! :slightly_smiling_face: 

To get rid of redundant code I suggest to merge the two dialogs ("LXQt Theme" and "Widget Style") and just have one "LXQt Theme" dialog. Otherwise we introduce code redundancy - in the first place because there is no "common interface" to properly handle the palette settings.

Probably a "mockup" is helpful:

LXQt_Appearance

tsujan:
> I suggest to merge the two dialogs ("LXQt Theme" and "Widget Style") 

No! Frankly, that's a bad idea.

antis81:
Sorry I don't have any more time to fix this.

PR #722, opened 2021-03-07
lxqt-config-appearance works in Wayland server.

selairi:
Requires: [https://github.com/lxqt/liblxqt/pull/286](https://github.com/lxqt/liblxqt/pull/286)
Gtk 3 needs gsettings instead xsettings in Wayland server.

### DIFF
--- a/liblxqt-config-cursor/crtheme.cpp
+++ b/liblxqt-config-cursor/crtheme.cpp
@@ -225,28 +225,28 @@ bool haveXfixes()
     bool result = false;
     int event_base, error_base;
 
-
-    if(QGuiApplication::platformName() == QStringLiteral("xcb"))
-        if (XFixesQueryExtension(QX11Info::display(), &event_base, &error_base))
-        {
-            int major, minor;
-            XFixesQueryVersion(QX11Info::display(), &major, &minor);
-            result = (major >= 2);
-        }
+    if (XFixesQueryExtension(QX11Info::display(), &event_base, &error_base))
+    {
+        int major, minor;
+        XFixesQueryVersion(QX11Info::display(), &major, &minor);
+        result = (major >= 2);
+    }
     return result;
 }
 
 bool applyTheme(const XCursorThemeData &theme, int cursorSize)
 {
-    // Require the Xcursor version that shipped with X11R6.9 or greater, since
-    // in previous versions the Xfixes code wasn't enabled due to a bug in the
-    // build system (freedesktop bug #975).
-    if (!haveXfixes()) return false;
-    
-    // Sets default cursor size
-    if(QGuiApplication::platformName() == QStringLiteral("xcb"))
-        XcursorSetDefaultSize(QX11Info::display(), cursorSize);
+    bool isX11 = QGuiApplication::platformName() == QStringLiteral("xcb"); 
+    if(isX11)
+    {
+        // Require the Xcursor version that shipped with X11R6.9 or greater, since
+        // in previous versions the Xfixes code wasn't enabled due to a bug in the
+        // build system (freedesktop bug #975).
+        if (!haveXfixes()) return false;
 
+        // Sets default cursor size
+        XcursorSetDefaultSize(QX11Info::display(), cursorSize);
+    }
     // Set up the proper launch environment for newly started apps
     //k8:!!!:KToolInvocation::klauncher()->setLaunchEnv("XCURSOR_THEME", themeName);
 
@@ -275,8 +275,7 @@ bool applyTheme(const XCursorThemeData &theme, int cursorSize)
     for (const QString &name : qAsConst(names))
     {
         Cursor cursor = (Cursor)theme.loadCursorHandle(name);
-
-        if(QGuiApplication::platformName() == QStringLiteral("xcb"))
+        if(isX11)
             XFixesChangeCursorByName(QX11Info::display(), cursor, QFile::encodeName(name).constData());
         // FIXME: do we need to free the cursor?
     }
--- a/liblxqt-config-cursor/crtheme.cpp
+++ b/liblxqt-config-cursor/crtheme.cpp
@@ -22,6 +22,7 @@
 #include <QDebug>
 
 #include "crtheme.h"
+#include "xcr/xcrimg.h"
 
 #include <QStyle>
 #include <QX11Info>
@@ -177,7 +178,8 @@ XcursorImages *XCursorThemeData::xcLoadImages(const QString &image, int size) co
 
 unsigned long XCursorThemeData::loadCursorHandle(const QString &name, int size) const
 {
-    if (size == -1) size = XcursorGetDefaultSize(QX11Info::display());
+    if (size == -1) size = getDefaultCursorSize();
+
     // Load the cursor images
     XcursorImages *images = xcLoadImages(name, size);
     if (!images) images = xcLoadImages(findAlternative(name), size);
@@ -185,7 +187,10 @@ unsigned long XCursorThemeData::loadCursorHandle(const QString &name, int size) 
     //if (!images) return LegacyTheme::loadCursor(name);
     if (!images) return 0;
     // Create the cursor
-    unsigned long handle = (unsigned long)XcursorImagesLoadCursor(QX11Info::display(), images);
+
+    unsigned long handle = 0;
+    if(QGuiApplication::platformName() == QStringLiteral("xcb"))
+        handle = (unsigned long)XcursorImagesLoadCursor(QX11Info::display(), images);
     XcursorImagesDestroy(images);
     //setCursorName(cursor, name);
     return handle;
@@ -193,7 +198,8 @@ unsigned long XCursorThemeData::loadCursorHandle(const QString &name, int size) 
 
 QImage XCursorThemeData::loadImage(const QString &name, int size) const
 {
-    if (size == -1) size = XcursorGetDefaultSize(QX11Info::display());
+    if (size == -1) size = getDefaultCursorSize();
+
     // Load the image
     XcursorImage *xcimage = xcLoadImage(name, size);
     if (!xcimage) xcimage = xcLoadImage(findAlternative(name), size);
@@ -218,12 +224,15 @@ bool haveXfixes()
 {
     bool result = false;
     int event_base, error_base;
-    if (XFixesQueryExtension(QX11Info::display(), &event_base, &error_base))
-    {
-        int major, minor;
-        XFixesQueryVersion(QX11Info::display(), &major, &minor);
-        result = (major >= 2);
-    }
+
+
+    if(QGuiApplication::platformName() == QStringLiteral("xcb"))
+        if (XFixesQueryExtension(QX11Info::display(), &event_base, &error_base))
+        {
+            int major, minor;
+            XFixesQueryVersion(QX11Info::display(), &major, &minor);
+            result = (major >= 2);
+        }
     return result;
 }
 
@@ -235,7 +244,8 @@ bool applyTheme(const XCursorThemeData &theme, int cursorSize)
     if (!haveXfixes()) return false;
     
     // Sets default cursor size
-    XcursorSetDefaultSize(QX11Info::display(), cursorSize);
+    if(QGuiApplication::platformName() == QStringLiteral("xcb"))
+        XcursorSetDefaultSize(QX11Info::display(), cursorSize);
 
     // Set up the proper launch environment for newly started apps
     //k8:!!!:KToolInvocation::klauncher()->setLaunchEnv("XCURSOR_THEME", themeName);
@@ -265,7 +275,9 @@ bool applyTheme(const XCursorThemeData &theme, int cursorSize)
     for (const QString &name : qAsConst(names))
     {
         Cursor cursor = (Cursor)theme.loadCursorHandle(name);
-        XFixesChangeCursorByName(QX11Info::display(), cursor, QFile::encodeName(name).constData());
+
+        if(QGuiApplication::platformName() == QStringLiteral("xcb"))
+            XFixesChangeCursorByName(QX11Info::display(), cursor, QFile::encodeName(name).constData());
         // FIXME: do we need to free the cursor?
     }
     return true;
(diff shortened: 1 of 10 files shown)


luis-pereira:
I'm currently not using GTK3 nor Wayland. I'm not the best reviewer to this PR.

tsujan:
I might not find time to test it under Weston soon but it'll be good to have it in 0.17.

@stefonarch, could you test the PR for setting GTK3 appearance under LXQt (X11)? If not, just tell me.

tsujan:
BTW, it should NOT require https://github.com/lxqt/liblxqt/pull/286 anymore.

stefonarch:
> BTW, it should NOT require [lxqt/liblxqt#286](https://github.com/lxqt/liblxqt/pull/286) anymore.

No, it _does_ need it, otherwise
```
...
[ 95%] Building CXX object src/CMakeFiles/lxqt-config.dir/mainwindow.cpp.o
[ 95%] Building CXX object lxqt-config-appearance/CMakeFiles/lxqt-config-appearance.dir/main.cpp.o
/home/stef/git/lxqt-config-gtk3/lxqt-config-appearance/main.cpp:33:10: fatal error: LXQt/lxqtplatform.h: File o directory non esistente
   33 | #include <LXQt/lxqtplatform.h>
      |          ^~~~~~~~~~~~~~~~~~~~~
```
Will make some tests later today.

lxqt-config-appearance/main.cpp
```diff
@@ -30,6 +30,8 @@
 #include <LXQt/Settings>
 #include <LXQt/ConfigDialog>
 #include <QCommandLineParser>
+#include <LXQt/lxqtplatform.h>
+
```
tsujan, line 34: Please remove the above header.

tsujan:
@stefonarch, it can be compiled without that line.

stefonarch:
Maybe I'm not the right person to test this, but I see that only gtk2 settings (for example breeze-dark and gnome-disk-utility) are applied, gtk3 settings had no effect. But it was like that also before.

tsujan:
You may have found a problem.

tsujan:
BTW, for testing GTK3 settings, `gtk3-widget-factory` is the best and simplest tool. (In very old times, I made GTK3 themes and tested them with it; my themes were broken by Gnome devs so many times and I stopped using GTK3.)

selairi:
Hi, I have just updated the code:
- GSettings is used instead xsettings. I added more GTK options and commented deprecated ones.
- lxqtplatform.h has been removed.
- Cursor themes are applied on Wayland, but restart is required. 
On Wayland, Qt reads XCURSOR_THEME environment variable at start or use default cursor (see file qwaylandinputdevice.cpp from qtwayland module). I can not find a way to update the cursor theme on Qt.

I have added a warning dialog, when restart is required. I have found that restart is required on X11, also. If you change cursor theme on X11, you will find that some windows don't update the cursor theme, example, the Openbox title bar.

Thanks

lxqt-config-appearance/main.cpp
selairi: Ok

selairi:
@stefonarch

> Maybe I'm not the right person to test this, but I see that only gtk2 settings (for example breeze-dark and gnome-disk-utility) are applied, gtk3 settings had no effect. But it was like that also before.

On X11 xsettingsd is used to update themes. xsettingsd can not apply different settings to GTK2 and GTK3. The GTK2 settings are applied by default. If you restart the LXQt session, the themes will be Ok.

tsujan (changes requested):
Worked under Weston after manual editing. The GTK theme didn't change on the fly but I don't think we need to worry about that. The user should restart his GTK apps to see the effect and, IMO, that's enough. But:

 1. It cannot be compiled in its current form because `#include <LXQt/lxqtplatform.h>` is still there.
 2. I don't think consulting `gsettings` under Wayland is a good idea for getting the cursor theme. `gsettings` is for Gnome, not for us. I don't have another solution but I think we shouldn't rely on `gsettings` here.

selairi:
@tsujan 
> Worked under Weston after manual editing. The GTK theme didn't change on the fly but I don't think we need to worry about that. The user should restart his GTK apps to see the effect and, IMO, that's enough.

You have to set the environment variables for LXQt and Qt theme changes should work. GTK3 theme should be changed on the fly, too. Qt cursor theme is not changed on the fly, because Qt library reads the cursor theme on start.

Please, try:

````
#!/bin/sh

contains()
{
    local str="$1" substr="$2"
    [ "$str" = "$substr" -o -z "${str##$substr:*}" -o -z "${str##*:$substr:*}" -o -z "${str%%*:$substr}" ]
}

if [ -z "$XDG_DATA_HOME" ]; then
    export XDG_DATA_HOME="$HOME/.local/share"
fi

if [ -z "$XDG_CONFIG_HOME" ]; then
    export XDG_CONFIG_HOME="$HOME/.config"
fi

if [ -z "$XDG_DATA_DIRS" ]; then
    XDG_DATA_DIRS="$XDG_DATA_HOME:/usr/local/share:/usr/share"
else
    if ! contains "$XDG_DATA_DIRS" "/usr/local/share"; then
        XDG_DATA_DIRS="$XDG_DATA_DIRS:/usr/local/share"
    fi
fi
export XDG_DATA_DIRS

if [ -z "$XDG_CONFIG_DIRS" ]; then
    export XDG_CONFIG_DIRS="/etc:/etc/xdg:/usr/share"
else
    if ! contains "$XDG_CONFIG_DIRS" '/etc/xdg'; then
        XDG_CONFIG_DIRS="$XDG_CONFIG_DIRS:/etc/xdg"
    fi
fi

if [ -z "$XDG_CACHE_HOME" ]; then
    export XDG_CACHE_HOME="$HOME/.cache"
fi

# Ensure the existence of the 'Desktop' folder
if [ -e "$XDG_CONFIG_HOME/user-dirs.dirs" ]; then
    . "$XDG_CONFIG_HOME/user-dirs.dirs"
else
    XDG_DESKTOP_DIR="$HOME/Desktop"
fi
mkdir -p "$XDG_DESKTOP_DIR"

export XDG_MENU_PREFIX="lxqt-"

export XDG_CURRENT_DESKTOP="LXQt"

export QTXDG_START_DETACH_TRULY=0

export QT_QPA_PLATFORM=wayland
export GDK_BACKEND=wayland
export QT_STYLE_OVERRIDE=fusion
export SAL_USE_VCLPLUGIN=gtk3
export QT_PLATFORM_PLUGIN=lxqt
export QT_QPA_PLATFORMTHEME=lxqt
export MOZ_ENABLE_WAYLAND=1
# Launch DBus if needed
if [ -z "$DBUS_SESSION_BUS_ADDRESS" ]; then
    if [ -z "$XDG_RUNTIME_DIR" ] || ! [ -S "$XDG_RUNTIME_DIR/bus" ] || ! [ -O "$XDG_RUNTIME_DIR/bus" ]; then
        eval "$(dbus-launch --sh-syntax --exit-with-session)" || echo "startlxqt: error executing dbus-launch" >&2
    fi
fi
# Uncomment your Wayland server
weston
# sway
# wayfire
````

> 
>     1. It cannot be compiled in its current form because `#include <LXQt/lxqtplatform.h>` is still there.
> 

Removed. Sorry.

>     2. I don't think consulting `gsettings` under Wayland is a good idea for getting the cursor theme. `gsettings` is for Gnome, not for us. I don't have another solution but I think we shouldn't rely on `gsettings` here.

This is a good point. gsettings should be used be lxqt-config-appearance in order to set and get GTK settings. GTK3 doesn't read its own config files in Wayland session. GTK3 reads gsettings values. Get default mouse theme in Wayland is easy with gsettings, then I wrote the code.

If gsettings is not used, then a function to read the cursor theme from XDG spec must be written:
[https://wiki.archlinux.org/index.php/Cursor_themes#XDG_specification](https://wiki.archlinux.org/index.php/Cursor_themes#XDG_specification)
"/home/user/.icons/default/index.theme" file must be parsed. If "/home/user/.icons/default/index.theme" doesn't exist, then "/usr/share/icons/default/index.theme" must be parsed.
I could write this function if you want.

Thanks for your suggestions.

**Comment by tsujan** [comment]
> gsettings should be used be lxqt-config-appearance in order to set and get GTK settings.

Yes, that's reasonable.

> "/home/user/.icons/default/index.theme" file must be parsed....

`lxqt-config-appearance` already does that; doesn't it? I have `~/.icons/default/index.theme`; it starts with: "# Written by lxqt-config-appearance".

**Comment by selairi** [comment]
> > "/home/user/.icons/default/index.theme" file must be parsed....
> 
> `lxqt-config-appearance` already does that; doesn't it? I have `~/.icons/default/index.theme`; it starts with: "# Written by lxqt-config-appearance".

Yes and no. As far as I know `~/.icons/default/index.theme` is not parsed, it is written by lxqt-config-appearance. The default cursor theme and size is read from X11 calls. There are lots of `index.theme` parsed by liblxqt-config-cursor. It is reasonable to parse `~/.icons/default/index.theme`, too... **Done**

The liblxqt-config-cursor/cfgfile.cpp parses `index.theme` files. It could be done by QSettings. The loadCfgFile function should be removed and use QSettings instead. But I think that it must be in another pull request.

**Comment by tsujan** [comment]
> The liblxqt-config-cursor/cfgfile.cpp parses index.theme files. It could be done by QSettings.

Maybe, when that function was written, `QSettings` couldn't handle spaces in keys or values. As far as I tested, it can in Linux.

Or there was another reason?

**Comment by selairi** [comment]
> > The liblxqt-config-cursor/cfgfile.cpp parses index.theme files. It could be done by QSettings.
> 
> Maybe, when that function was written, `QSettings` couldn't handle spaces in keys or values. As far as I tested, it can in Linux.
 
The problems one by one, please. First, I would prefer merge the actual PR. Then, a new issue should be opened about QSettings and liblxqt-config-cursor.

**Comment by tsujan** [comment]
> The problems one by one, please.

I didn't mean it should be fixed/changed right now or even before the next release.

I'll review the recent changes to this PR tonight.

File: liblxqt-config-cursor/crtheme.cpp [hunk 2]
```diff
@@ -265,13 +275,33 @@ bool applyTheme(const XCursorThemeData &theme, int cursorSize)
     for (const QString &name : qAsConst(names))
     {
         Cursor cursor = (Cursor)theme.loadCursorHandle(name);
-        XFixesChangeCursorByName(QX11Info::display(), cursor, QFile::encodeName(name).constData());
+
+        if(QGuiApplication::platformName() == QStringLiteral("xcb"))
+            XFixesChangeCursorByName(QX11Info::display(), cursor, QFile::encodeName(name).constData());
```
tsujan, line 280: in `applyTheme()`, there are too many xcb checks (especially inside the loop).

If your goal is returning `true` from this function when the platform is not xcb, IMHO, it's better to do it once, before the comment `// Sets default cursor size`:

```c++
    if(QGuiApplication::platformName() != QStringLiteral("xcb"))
        return true;
```

lxqt-config-appearance/configothertoolkits.cpp
```diff
@@ -61,20 +64,39 @@ gtk-menu-images = %4
...
+)GTK3_GSETTINGS";
+
+/*
+ * Button and menu images are deprecated in GTK 3.10:
+ *    https://developer.gnome.org/gtk3/stable/GtkSettings.html#GtkSettings--gtk-menu-images
+ * Toolbar-style is also deprecated.
+ * GTK just doesn't intend to support image menu items. They're considered bad practice in modern UI. 
+ */
+
+ConfigOtherToolKits::ConfigOtherToolKits(LXQt::Settings *settings,  LXQt::Settings *configAppearanceSettings, QWidget *parent) : QObject(parent)
```
tsujan, line 99: Would you please tell why you changed `QObject *parent` to `QWidget *parent`?

lxqt-config-appearance/configothertoolkits.cpp
```diff
@@ -189,6 +218,24 @@ void ConfigOtherToolKits::setXSettingsConfig()
...
+    QString commands;
+    if(version == QLatin1String("3.0"))
+        commands = getConfig(GTK3_GSETTINGS, ConfigOtherToolKits::GTK3GSETTINGS);
+    for(QString command : commands.split(QStringLiteral("\n"))) {
+        if(command.isEmpty() or command.startsWith(QStringLiteral("#")))
+            continue;
+        QStringList args = QProcess::splitCommand(QStringView(command));
+        bool ok = QProcess::startDetached(QStringLiteral("gsettings"), args);
+        if(!ok)
+            QMessageBox::critical((QWidget*) parent(), tr("Error"), tr("Error: gsettings can not be run"));
```
tsujan, line 235: 1. If `gsettings` is not found, the error message needs to be shown only once and the loop should be ended. Here, it's shown with each loop cycle.
 2. A small typo: it's "Error: gsettings cannot be run" (you wrote "can not").

lxqt-config-appearance/main.cpp
```diff
@@ -104,21 +107,27 @@ int main (int argc, char **argv)
...
     cursorPage->setCurrent();
     dialog->addPage(cursorPage, QObject::tr("Cursor"), QStringList() << QStringLiteral("input-mouse") << QStringLiteral("preferences-desktop"));
-    QObject::connect(cursorPage, &SelectWnd::settingsChanged, dialog, [dialog] {
-        dialog->enableButton(QDialogButtonBox::Apply, true);
-    });
+    std::shared_ptr<bool> cursorChanged = std::make_shared<bool>(false);
+    QObject::connect(cursorPage, &SelectWnd::settingsChanged, dialog, [dialog, cursorChanged] {
+            dialog->enableButton(QDialogButtonBox::Apply, true);
+            *cursorChanged = true;
+            });
```
tsujan, line 114: I'm not sure whether it's necessary to show a message but, in case it is:

Instead of this, wouldn't it be better if `SelectWnd` had a small method to tell whether the cursor was changed or not? I think a private boolean and a public method that gets it would be enough.

tsujan: OK, I tested under X11 (our main goal) and the cursor was changed on the fly. The message isn't needed.

If it's needed under Wayland, please limit it to Wayland by adding the above-mentioned method to `SelectWnd`. However, I think it's better not to show any message under Wayland either.

liblxqt-config-cursor/crtheme.cpp
selairi: OK

lxqt-config-appearance/configothertoolkits.cpp
selairi: Sorry. I changed but I don't remember why. QObject is better than QWidget here. I will restore QObject.

lxqt-config-appearance/configothertoolkits.cpp
selairi: Ok. Fixed.

lxqt-config-appearance/main.cpp
selairi: > OK, I tested under X11 (our main target) and the cursor was changed on the fly. The message isn't needed.

Open a LXQt session with openbox as window manager. Then change the cursor style. Finally, put your cursor over an openbox title bar. Openbox keeps the previous cursor theme.
I have installed mousepad in my computer to test GTK3 themes, and mousepad doesn't change cursor theme.

I have just found this: If you see on top of SelectWnd, there is this text: "LXQt needs restart after this change". I will remove the dialog.

Thanks for your review.

tsujan (approved):
GTM

lxqt-config-appearance/main.cpp
tsujan: > Openbox keeps the previous cursor theme.

I tested with KWin; I don't use Openbox. But you're right.

Issue #734, opened 2021-04-15
Fails to build with Qt5 5.12.10


darkshram:
This is another `QStringView` issue:
```
[100%] Building CXX object lxqt-config-appearance/CMakeFiles/lxqt-config-appearance.dir/LXQtAppTranslationLoader.cpp.o
[...]

PR #735, opened 2021-04-15
Fixed compilation against Q < 5.15

tsujan:
Closes https://github.com/lxqt/lxqt-config/issues/734


tsujan:
Reported and tested by @darkshram. Merging....

NOTE: At most, the new Wayland code might show an error message about not finding `gsettings` with Qt < 5.15 and under Wayland (although I think it won't) but those with Qt < 5.15 won't need it.

PR #762, opened 2021-08-30
Do not force openbox configs under LXQt

stefonarch:
Twin of https://github.com/lxqt/lxqt-panel/pull/1656


tsujan (approved):
GTM

tsujan:
Some users might complain but I think it's the correct thing to do.

stefonarch:
I was thinking about an WM menu in the settings, but as the default names of the config tools are not clear that's even worse,

tsujan:
Being WM-agnostic is a smart thing in LXQt. Users can easily find the settings of their chosen WM.

Issue #761, opened 2021-08-30
Control Center doesn't display everything


stefonarch:
##### Expected Behavior

All items present in Mainmenu > Preferences are also displayed in LXQt Control Center.
##### Current Behavior

Some are missing: kde-systemsettings,ibus settings, advanced network configuration and else
##### Possible Solution

Fix `/etc/menus/lxqt-config.menu` which has a menu for "others" which is
* never displayed (apparently not working)
* refers to an inexistent `lxqt-settings-other.directory` file - existing is only` lxde-others.directory`
The name should also be more explicit: lxqt-control-center.menu as it's possibile to chose it in mainmenu settings and doesn't give a good UX.

##### Context

Control Center could be the only configuration tool in a menu, but this has to be fixed.
##### System Information

* Distribution & Version: all recent

tsujan:
I ask these questions from myself: lxqt-panel is an app; why should the settings of one of its widgets be in Control Center? Do you really want to clutter Control Center with every setting we find? Why should we put all config tools in it, while the main menu has sections for them?...

tsujan:
> Fix /etc/menus/lxqt-config.menu which has a menu for "others"

"Others" is shown here.

stefonarch:
I've never seen it, neither in installations nor in VM's as I remember.
What do you have in "others"? Are ibus settings and im-config visible there?

> Do you really want to clutter Control Center with every setting we find? Why should we put all config tools in it, while the main menu has sections for them?

When we have something called "Configuration Center" it should be complete. And cluttering is not an issue IMHO, with three sections. As said in https://github.com/lxqt/lxqt-config/issues/759#issuecomment-907817973, I'd like to offer more menu files to chose from, one would have only the configuration center but it needs to be complete.

Draft:

![schermata-08-30-17-10](https://user-images.githubusercontent.com/10681413/131362123-034893e0-3b5a-4059-bc6f-3d136e3effdf.png)

tsujan:
> What do you have in "others"?

I'll check later. It's in my other computer, where I have lots of garbage. "Others" is usually for garbage ;)

> When we have something called "Configuration Center" it should be complete.

It's *LXQt's* config center plus system settings. Why, for example, should we have KDE's system settings there? That would mislead users: they change font settings in KDE's settings and see no change in LXQt. Seeing KDE's keyboard shortcuts there is even worse.

tsujan:
[...]

PR #764, opened 2021-09-03
Make 'other settings' work in Configuration Center

stefonarch:
It was half implemented, .directory and translation files missing. 
Will not show kdesystemsettings. 

schermata-09-03-16-55

Closes https://github.com/lxqt/lxqt-config/issues/761

### DIFF
--- a/src/lxqt-config.menu
+++ b/src/lxqt-config.menu
@@ -45,12 +45,12 @@
     <Menu>
         <Name>Other settings</Name>
         <Directory>lxqt-settings-other.directory</Directory>
-        <OnlyUnallocated />
         <Include>
             <And>
                 <Category>Settings</Category>
                 <Not>
                     <Or>
+                        <Filename>pcmanfm-qt-desktop-pref.desktop</Filename>
                         <Category>LXQt</Category>
                         <Category>LXDE</Category>
                         <Category>System</Category>
@@ -63,6 +63,7 @@
     <Layout>
         <Menuname>LXQt settings</Menuname>
         <Menuname>System settings</Menuname>
+        <Menuname>Other settings</Menuname>
         <Merge type="menus"/>
     </Layout>
 </Menu>
--- a/src/menuname/desktop-directories/lxqt-settings-other.directory.in
+++ b/src/menuname/desktop-directories/lxqt-settings-other.directory.in
@@ -1,0 +1,3 @@
+[Desktop Entry]
+Type=Directory
+
--- a/src/menuname/translations/lxqt-settings-other.directory.yaml
+++ b/src/menuname/translations/lxqt-settings-other.directory.yaml
@@ -1,0 +1,1 @@
+Desktop Entry/Name: "Other Settings"


src/lxqt-config.menu
```diff
@@ -45,12 +45,13 @@
...
         <Name>Other settings</Name>
         <Directory>lxqt-settings-other.directory</Directory>
-        <OnlyUnallocated />
         <Include>
             <And>
                 <Category>Settings</Category>
                 <Not>
                     <Or>
+                        <Filename>kdesystemsettings.desktop</Filename>
+                        <Filename>pcmanfm-qt-desktop-pref.desktop</Filename>
```
tsujan, line 54: Very sorry, I saw the PR only now.

This section seems to be about exclusion. What about this instead of `<Filename>kdesystemsettings.desktop</Filename>`?

```xml
<Category>KDE</Category>
<Category>GNOME</Category>
```

And what about other DE categories? I think "Other settings" entries were left out to be decided later. Maybe, it's time to make a decision.

stefonarch: What should be shown are  the (once removed `OnlyShowIn=XFCE;`)  three Xfwm4 settings. 
KDE and Gnome could be excluded, not sure about  others, probably yes.

tsujan: > What should be shown are the three Xfwm4 settings.

We can make exceptions for them.

I'll play with it later. It'll be good if we decide about it before the next release.

tsujan:
After playing with various lines, I think it's better just to remove `<Filename>kdesystemsettings.desktop</Filename>` from it and merge it:

```xml
    <Menu>
        <Name>Other settings</Name>
        <Directory>lxqt-settings-other.directory</Directory>
        <Include>
            <And>
                <Category>Settings</Category>
                <Not>
                    <Or>
                        <Filename>pcmanfm-qt-desktop-pref.desktop</Filename>
                        <Category>LXQt</Category>
                        <Category>LXDE</Category>
                        <Category>System</Category>
                    </Or>
                </Not>
            </And>
         </Include>
     </Menu>

    <Layout>
        <Menuname>LXQt settings</Menuname>
        <Menuname>System settings</Menuname>
        <Menuname>Other settings</Menuname>
        <Merge type="menus"/>
    </Layout>
```

Of course, kdesystemsettings appears in it but, since it's under "Other Settings", there's no problem.

If we think of "Other Settings" as a section for all desktop files belonging to `Settings` category but not to LXQt or `System`, it's logical to have kdesystemsettings and the settings for other DEs under it, especially when it's the last section.

It's almost what you'd proposed at https://github.com/lxqt/lxqt-config/issues/761#issuecomment-908428350. On that page, I didn't know that, if we excluded DEs, we'd lose some good settings other than kdesystemsettings (like Firewall Configuration and Bluetooth Manager); hence, my change of mind here.

tsujan:
BTW, do you know why "s" isn't capitalized in "Other settings"?

stefonarch:
Here it is capitalized, and translation is working too, did just a test.
https://github.com/stefonarch/lxqt-config/blob/show_other_settings/src/menuname/translations/lxqt-settings-other.directory.yaml
schermata-10-05-11-58

tsujan (approved):
IMO, it's good.

> Here it is capitalized

I didn't install it; just changed the menu. Perhaps that's the cause.

tsujan:
I installed it and, yes, it's capitalized now :)

capitalized

PR #769, opened 2021-09-29
Replace QLatin1String with QStringLiteral

iamthesenate1:
Output of clavy with CLAZY_CHECKS="qstring-allocations".

I've seen that opinions are divided about which one is better.
From my point of view, if this PR is merged, good. If not, also good.

### DIFF
--- a/liblxqt-config-cursor/cfgfile.cpp
+++ b/liblxqt-config-cursor/cfgfile.cpp
@@ -114,7 +114,7 @@ const QString findDefaultTheme()
             if (!s.startsWith(QLatin1String("Xcursor.theme:"))) continue;
             s.remove(0, 14);
             s = s.trimmed();
-            if (s.isEmpty()) s = QLatin1String("default");
+            if (s.isEmpty()) s = QStringLiteral("default");
             res = s;
         }
         fl.close();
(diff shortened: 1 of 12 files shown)


tsujan:
Personally, I think `QStringLiteral` is better in these places.

> I've seen that opinions are divided about which one is better.

Yes, I've noticed it too.

Contrary to what Qt doc implies, I've seen no tangible gain in using either of them instead of `QString::QString(const char *str)` for (tens of) thousands of strings (while I've seen big gains in quite different things, about which Qt doc is silent).

tsujan (changes requested):
On second thought and apart from personal opinions (including mine), if we want to follow Qt doc, we shouldn't change *all* of these cases. According to [Qt doc](https://doc.qt.io/qt-5/qstring.html#QStringLiteral):

> Note: QLatin1String can still be more efficient than QStringLiteral when the string is passed
> to a function that has an overload taking QLatin1String and this overload avoids conversion to QString. 

`QString::operator=`, `QString::operator==` and `QString::operator+=` are among those overloads.

But, for example, `QSettings::setValue` doesn't have such an overload and `QStringLiteral` seems a better choice for it.

All in all, changes may be made but more carefully.

iamthesenate1:
I might check for such cases later. For now, I'm very busy and I regard these changes as a pretty low priority.

tsujan:
Take your time. There's no rush at all.

Issue #771, opened 2021-10-01
[Feature Request] Default dark palette


stefonarch:
Add  a second button for  "default dark palette" in lxqt-config-appearance.

##### Current Behavior

Setting a dark palette needs many clicks atm, switching from dark to bright theming and back would be easier.
##### Possible Solution

As above

##### Context

In theme testing it's annoying always to set at least 4 colours by hand.

tsujan:
Won't it be better if we add a way of saving/restoring user-defined palettes instead?

stefonarch:
Hm, why not both?  :)
But for first time users of palette a dark default one would be good IMHO, so users can easily switch. 
Not sure but could the dark palette be triggered also be selecting dark themes in some way?

tsujan:
> why not both?

Because making a dark palette is a piece of cake and the second functionality makes the first one redundant. Moreover, there's really no default dark palette in Qt, while Qt does have a default palette, which is light.

tsujan:
> Not sure but could the dark palette be triggered also be selecting dark themes in some way?

You mean LXQt (panel) themes. They're in QSS, which is at odds with QStyle. We can't know if an LXQt theme is light or dark.

But, if palettes are saveable and restorable, palettes with the same names could be added and applied.

stefonarch:
> But, if palettes are saveable and restorable, palettes with the same names could be added and applied.

That sounds good, and there would be any kind of palette then?

> > why not both?
> 
> Because making a dark palette is a piece of cake and the second functionality makes the first one redundant. Moreover, there's really no default dark palette in Qt, while Qt does have a default palette, which is light.

Right, I don't have those insights in Qt.

tsujan:
> and there would be any kind of palette then?

Why not? We could add palettes corresponding to our LXQt themes, and more.

Its logic is simple: like KDE color schemes but without style-specific components. The implementation will take time but I'm quite motivated (although I won't use this feature -- I use Kvantum, which is independent of palettes).

tsujan:
This is what I mean:

![palettes](https://user-images.githubusercontent.com/981076/135751643-b972cc59-4325-4b86-be30-a395c2bda114.png)

When "Save Palette" is clicked, this dialog will appear:

![save_palette](https://user-images.githubusercontent.com/981076/135751660-d3ac065a-ca2f-4829-8044-bf014067d512.png)

And this one shows up if "Load Palette" is clicked and there are palettes:

[...]

PR #772, opened 2021-10-04
Save and load Qt palettes

tsujan:
Palettes are saved into `$XDG_DATA_HOME/lxqt/palettes/` and loaded from `$XDG_DATA_DIRS/lxqt/palettes/` (root palettes can be installed in places like `/usr/share/lxqt/palettes/`). The patch only relies on Qt to handle these directories.

Palette files have no extension; they're distinguished by their parent folders and their names.

Malformed files or sections are silently ignored. Although they're listed, the active palette is used instead of them.

NOTE: In this PR, I also corrected an old mistake (typo) that, by chance, had no effect.

Closes https://github.com/lxqt/lxqt-config/issues/771

### DIFF
--- a/lxqt-config-appearance/CMakeLists.txt
+++ b/lxqt-config-appearance/CMakeLists.txt
@@ -34,6 +34,7 @@ set(UI_FILES
     iconthemeconfig.ui
     lxqtthemeconfig.ui
     fontsconfig.ui
+    palettes.ui
     styleconfig.ui
 )
 
--- a/lxqt-config-appearance/palettes.ui
+++ b/lxqt-config-appearance/palettes.ui
@@ -1,0 +1,94 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>PalettesDialog</class>
+ <widget class="QDialog" name="Dialog">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>350</width>
+    <height>400</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Palettes</string>
+  </property>
+  <layout class="QGridLayout" name="gridLayout">
+   <item row="0" column="0" colspan="2">
+    <widget class="QLabel" name="label">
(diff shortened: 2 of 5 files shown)


stefonarch:
Works like  a charm!
 One small issue is the "Save palette" window which is not large enough so the title is truncated here.

schermata-10-04-13-08

Saving files with a space looks working, with a trailing slash at the end it results in an empty directory, but I think nobody will do that.

tsujan:
> the "Save palette" window which is not large enough

Sadly, it's what `QInputDialog` gives us and can't be changed easily. But I'll wrestle with it.

> with a trailing slash at the end it results in an empty directory

Interesting test! It didn't occur to me; will prevent it in the code.

Thanks!

stefonarch:
Or just make the title "Save"? This should fit well.

tsujan:
Added a line for preventing slash (and more).

As for input dialog, its width can be made larger by replacing a single and simple line of code with several lines. I think it's better to live with what Qt offers by default; changing it isn't worth cluttering the code.

stefonarch:
see my previous posting

tsujan:
> see my previous posting

I did; we added comments at the same time.

"Save" is short in English but who knows about other languages? IMHO, how the title appears isn't important. It's obvious what that dialog is for; the user couldn't mistake it for anything else.

tsujan:
BTW, later we could add palettes to some LXQt themes — palette names could be those of themes and they could be installed as a part of `lxqt-themes`. We could also add a mechanism to apply such palettes alongside their corresponding themes.

But, maybe, after the next release.

stefonarch:
> BTW, later we could add palettes to some LXQt themes

I tried some palettes for the themes, but I noticed that I don't like them in the same colors, to strong. Maybe just some "nice" default palettes?

tsujan:
> I tried some palettes for the themes, but I noticed that I don't like them

We'll have lots of time to find good palettes for LXQt themes. BTW, the palette you used in your screenshot looks nice.

tsujan:
Let's give it to git users for more tests.

Issue #801, opened 2021-12-11
Translation not displayed in "remove palette" dialog


stefonarch:
In https://github.com/lxqt/lxqt-config/blob/e4810a3bdacdf980b43c5cd0803f58f2aeb56f87/lxqt-config-appearance/styleconfig.cpp#L372
 ```
QMessageBox::StandardButton btn =
            QMessageBox::question(this,
                                  tr("Remove Palettes"),
                                  tr("Do you really want to remove selected palette(s)?\nRoot palettes will remain intact if existing."));
            if (btn != QMessageBox::Yes)
```
the translations exists (tested it and ru) but aren't displayed

##### Steps to Reproduce (for bugs)

1.  `LANG=it lxqt-config-appearance 
`
2. Open "carica schema" dialog
3. Click left button "rimuovi"
4. popup window appears in english

 Latest git and debian

tsujan:
Only "Remove Palettes"? The Palettes dialog itself is translated?

stefonarch:
Neither.
![schermata-12-12-19-15](https://user-images.githubusercontent.com/10681413/145724440-5139f9a4-ffa0-418b-9b99-aeac3c9c7cfb.png)

tsujan:
I tested and the problem was only in "Remove Palettes" dialog, not in "Palettes" dialog itself (that was what I wanted to know):

![tr](https://user-images.githubusercontent.com/981076/145727874-eb7d4fcf-3498-4f77-a252-b9a9716e78c8.png)

It may have something to do with `PalettesDialog` being a "local class". Will investigate it.

tsujan:
Yes, my guess was right. Once I put the translations outside the local class, they were shown:

![tr1](https://user-images.githubusercontent.com/981076/145728336-203f46d0-09f6-4d2b-8079-31243bb3d647.png)

Apparently, Qt's translation doesn't work inside local classes. Will make a PR.

PR #802, opened 2021-12-12
Fixed translation in "Remove Palettes" dialog

tsujan:
Qt doesn't translate inside local classes.

Fixes https://github.com/lxqt/lxqt-config/issues/801


stefonarch (approved):
Tested and working fine, GTM

stefonarch:
Wondering if there could be a better wording for "root palettes". "Default palettes"? "System palettes"?

tsujan:
> Wondering if there could be a better wording for "root palettes".

"Root" shows the contrast with "user". They're also installed as root. But feel free to change it later if you find a better word.

They aren't default palettes (we don't use them anywhere). The word "system" might make the user think that they're related to his system, while they aren't.

PR #804, opened 2021-12-23
Remove Qt4 dead conditional code

luis-pereira:
Q_WS_X11 was removed in Qt5. This code is dead since we ported to Qt5.


tsujan (approved):
GTM

Issue #820, opened 2022-03-12
Remember the size of lxqt-config-input dialog


ianfarrell47:
##### Issue
The default dimensions of lxqt-config-input's GUI should include more room for the page pane.
This is the dimensions that lxqt-config-input starts with:
![157998378-41415f0a-516d-4e42-b6d8-13c3105b6f54](https://user-images.githubusercontent.com/96850319/158002469-f29bd9d6-567e-4013-a02b-990501977196.png)

All pages fit well in the minimum dimensions  already, this would only increase the size so they have more room.

##### Possible solution
Increase minimum size of window to include more room. If DPI changes things too large, allow for scrollbar.

##### Drawbacks to solution
Code complexity and bigger dimensions on launch

##### System Information
lxqt-config-git - 1.0.0.25.g5ab3a64-1 5ab3a64
liblxqt - 1.0.0-1
lxqt-qtplugin - 1.0.0-2
qt5-base - 5.15.3+kde+r133-1
Qt5 style: Fusion
Font: 12 Noto Sans Display
linux 5.15.27-1-lts

tsujan:
There's no reason to give more width to the left pane. The width is calculated correctly.

tsujan:
You probably meant the height, not the width.

"LXQt Appearance Configuration" remembers its size but "Keyboard and Mouse Settings" dialog doesn't. We can't avoid scrollbars at the first run because different users may have different styles/fonts/... but we can remember the dialog size.

I'm changing the title and reopening this....

ianfarrell47:
Should the height be corrected to something more reasonable? I understand now that the width/height is dynamic (corresponding to the font size). The default window dimensions appear to be the same as the minimum dimensions. Perhaps we could increase minimum dimensions to a height more reasonable.

tsujan:
Searching for a "reasonable default height" isn't reasonable.

PR #821, opened 2022-03-12
Remember size of lxqt-config-input dialog

tsujan:
Closes https://github.com/lxqt/lxqt-config/issues/820

### DIFF
--- a/lxqt-config-input/lxqt-config-input.cpp
+++ b/lxqt-config-input/lxqt-config-input.cpp
@@ -136,10 +136,14 @@ int main(int argc, char** argv) {
 
     dlg.setWindowIcon(QIcon::fromTheme(QStringLiteral("input-keyboard")));
 
+    LXQt::Settings mConfigInputSettings(QStringLiteral("lxqt-config-input"));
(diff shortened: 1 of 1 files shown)


stefonarch:
GTM, and looks as it solved also a stupid behaviour from kwin: it open always in the upper left corner.

tsujan:
> looks as it solved also a stupid behaviour from kwin: it opened always in the upper left corner.

That has never happened here. But, recently, KWin's default behavior changed, especially its window positioning. I reversed it to the previous behavior as soon as I saw the change.

That being said, Qt always centers a dialog when it has no parent. If you saw that KWin put the dialog at another position under X11, you've encountered a regression in KWin.

tsujan:
BTW, I have a feeling that you may have forgotten https://github.com/lxqt/lxqt-panel/pull/1731 and https://github.com/lxqt/qterminal/pull/919 ;)

stefonarch:
The first I've forgotten, in qterminal I0m waiting for you :)
Will do this evening.

tsujan:
> in qterminal I'm waiting for you :)

Why? Have I forgotten something?

> Will do this evening.

There's no rush at all; just don't forget them.

Issue #834, opened 2022-04-01
Small GUI bug in LXQt Locale Configuration


tsujan:
##### Steps to Reproduce (for bugs)

1. Copy the exact text of your region from the Region combo-box.
2. Change the region by opening the combo list, so that something in the "Examples" section changes.
3. Paste the text from step 1 into the Region combo-box and either choose it from the completer or just press `Enter`.

Then the locale doesn't change (see the "Examples" section). You should use the mouse wheel inside the combo-box and change it one step forward and then backward again.

##### Possible Solution

Will fix it soon.

##### System Information

* Latest git LXQt

PR #835, opened 2022-04-01
Fixed manual editing of combos in LXQt Locale Config

tsujan:
Previously, manual editing only changed the text, not the locale.

Also made the GUI more up-to-date, by using group boxes, and removed redundant widgets from it.

Fixes https://github.com/lxqt/lxqt-config/issues/834


stefonarch (approved):
Didn't find anything not working. As previously said I never manage to see "no change" here.

tsujan:
> As previously said I never manage to see "no change" here.

I know but this isn't about "No change", which was OK.

PR #843, opened 2022-04-19
Add Greek (el) translation

dglent:
Add Greek [el] translation for desktop files 'Brightness' and 'Locale'


stefonarch:
Thanks, but all translations should go over weblate:
https://translate.lxqt-project.org/projects/lxqt-configuration/lxqt-config-brightness-menu-entry/
I saw that your permissions were not upgraded, I did it now so you can add a new translation file now (users have to ask for that).

PR #847, opened 2022-04-23
Consider LXQt theme names case-insensitively

tsujan:
This is needed after https://github.com/lxqt/liblxqt/pull/308

### DIFF
--- a/lxqt-config-appearance/lxqtthemeconfig.cpp
+++ b/lxqt-config-appearance/lxqtthemeconfig.cpp
@@ -113,7 +113,8 @@ void LXQtThemeConfig::initControls()
 
     QTreeWidgetItemIterator it(ui->lxqtThemeList);
     while (*it) {
-        if ((*it)->data(0, Qt::UserRole).toString() == currentTheme)
+        // liblxqt considers LXQt theme names case-insensitively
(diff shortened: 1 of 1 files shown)


tsujan:
This need changes, for two reasons: (1) `lxqt-config` shouldn't need to know about case-insensitivity; and (1) it shouldn't list theme variants whose differences are only in letter cases.

tsujan:
Oh, my memory failed me: (2) isn't a problem because `LXQt::LXQtTheme::allThemes()` is used.

Only knowing about case-insensitivity is a logical problem.

palinek:
I still think, the whole case insensitive comparison of theme directories is wrong concept.

tsujan:
Superseded by https://github.com/lxqt/lxqt-config/pull/848

tsujan:
> I still think, the whole case insensitive comparison of theme directories is wrong concept.

Yes, I know that you compare it with MS Windows file system. But it removes unnecessary complications.

tsujan:
BTW, while doing this work, I found a self-inconsistency in `liblxqt`. It treats theme paths and theme names on the same level when making a theme object but only uses the theme names in other places.

Fortunately, it doesn't affect any code in LXQt because, everywhere, the names are used for comparisons.

PR #848, opened 2022-04-24
Use real LXQt theme name for comparison

tsujan:
Previously, the value of the "theme" key was used, which may differ in letter cases from the current theme name.

Also, the themes list is sorted alphabetically. Previously, the user themes came first.

This PR requires and depends on https://github.com/lxqt/liblxqt/pull/309.


tsujan:
Closing. See https://github.com/lxqt/liblxqt/pull/310 for the reason.

PR #853, opened 2022-05-09
Improved name for `set wallpaper from theme`

stefonarch:


tsujan (approved):
GTM.

BTW, the old problem with the Hebrew translation is corrected here.

Issue #852, opened 2022-05-09
Appearance Config: Distinguish user themes and open theme folders


tsujan:
Inspired by my discussion with @stefonarch at the end of this page (the PR was reversed later): https://github.com/lxqt/liblxqt/pull/308

 1. Themes are always sorted alphabetically.
 2. User themes are distinguished by "(User Theme)" after their names.
 3. Double clicking a theme opens its folder.
 4. Also, each theme has a context menu for opening its folder.

A screenshot:

![themes](https://user-images.githubusercontent.com/981076/167445912-9c1425c4-2fd6-4538-af07-8c3f23f8de20.png)

stefonarch:
That looks fine!
BTW I was never happy with "Override user-defined wallpaper". What about changing it to "Use  wallpaperprovided by the theme" ?

tsujan:
Good idea. Would you make a PR for it?

stefonarch:
→ https://github.com/lxqt/lxqt-config/pull/853

ianfarrell47:
I'm looking forward to this in 1.1.1 or 1.2.0.

One other bit of polish on this panel that I see, for lack of a better spot to mention it, is to have consistent preview images. I can create an Issue and compile those images, and get it into 1.1.1 or 1.2.0, granted Git doesn't give me any issues.

stefonarch:
Nice idea! I was thinking (related to making themes translatable) having also a description, but at the end it's not so important as users see quite immediately how the theme is.

And there is a not so small part of users that never touch themes ;) - I know some of theme.

tsujan:
> I was thinking (related to making themes translatable) having also a description, but at the end it's not so important

I think we should do it before 1.1.2. Having a desktop entry for each theme has also this advantage that the theme name can be changed without changing its folder name, the latter being backward incompatible.

ianfarrell47:
The .desktop files I feel are an okay idea, my only confusion there is that .desktop is typically used for launching applications, whereas raw data is typically stored in an INI.

I assumed the preview images were PNGs but I do not see them in the repository.

tsujan:
> my only confusion there is that .desktop is typically used for launching applications

My mistake; sorry! It can be `index.theme`.

stefonarch:
> I assumed the preview images were PNGs but I do not see them in the repository.

https://github.com/lxqt/lxqt-themes/blob/master/themes/Arch-Colors/preview.png

[...]

PR #854, opened 2022-05-10
Small enhancement to theme list in appearance config dialog

tsujan:
Themes are always sorted alphabetically, user themes are distinguished, and theme folders can be opened by double clicking or by using context menus.

Closes https://github.com/lxqt/lxqt-config/issues/852


tsujan:
@stefonarch 
Would you please test it (again)? I made a small change.

PR #858, opened 2022-05-16
lxqt-config-brightness: Set minimum backlight

selairi:
In some hardware the 5% as minimum value of backlight can be too brightness. Advanced options has been added to select this minimum value:

### DIFF
--- a/lxqt-config-brightness/brightnesssettings.cpp
+++ b/lxqt-config-brightness/brightnesssettings.cpp
@@ -21,7 +21,7 @@
 #include <QMessageBox>
 #include <QPushButton>
 
-BrightnessSettings::BrightnessSettings(QWidget *parent):QDialog(parent)
+BrightnessSettings::BrightnessSettings(LXQt::Settings *settings, QWidget *parent):QDialog(parent)
 {
     ui = new Ui::BrightnessSettings();
     ui->setupUi(this);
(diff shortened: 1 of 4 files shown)


lxqt-config-brightness/brightnesssettings.ui
```diff
@@ -19,94 +19,181 @@
...
+       <item>
+        <widget class="QGroupBox" name="groupBox_2">
+         <property name="title">
+          <string>Backlight</string>
+         </property>
+         <layout class="QVBoxLayout" name="verticalLayout_5">
+          <item>
+           <widget class="QGroupBox" name="groupBox_3">
+            <property name="toolTip">
+             <string>The minimum value for backlight value can be really brightness.
```
tsujan, line 124: This sentence needs to be corrected. Did you mean, "The minimum value of backlight can be too hight"?

lxqt-config-brightness/brightnesssettings.ui
```diff
@@ -19,94 +19,181 @@
...
+         <property name="title">
+          <string>Backlight</string>
+         </property>
+         <layout class="QVBoxLayout" name="verticalLayout_5">
+          <item>
+           <widget class="QGroupBox" name="groupBox_3">
+            <property name="toolTip">
+             <string>The minimum value for backlight value can be really brightness. 
+It is possible set this minimum value.
+Warning: If a very low backlight is set, a black screen can be shown.</string>
```
tsujan, line 126: IMHO, the first sentence isn't needed, and this tooltip can be changed to something like:

The minimum value of backlight.
Warning: A very low value can result in a black screen.

lxqt-config-brightness/brightnesssettings.cpp
```diff
@@ -171,18 +210,10 @@ void BrightnessSettings::revertValues()
...
-        ui->backlightSlider->setMaximum(100);
-        // Set the minimum to 5% of the maximum to prevent a black screen
-        ui->backlightSlider->setMinimum(5);
-        ui->backlightSlider->setValue( (value * 100) / maxBacklight);
-    }
+
+    ui->backlightSlider->setMaximum(maxBacklight);
+    ui->backlightSlider->setMinimum(minBacklight);
+    ui->backlightSlider->setValue(value);
 }
```
tsujan, line 219: A newline is missing here.

tsujan:
IMHO, the code needs to take care of the minimum value automatically. The new GUI seems too confusing to me.

Do other DEs do this?

lxqt-config-brightness/brightnesssettings.ui
selairi: Ok. Fixed.

selairi:
Do other DEs do this?

No. Other DEs doesn't set a minimum value for the backlight. For example, I have downloaded the code of XFCE and I have checked the minimum value for the backlight and it is set to 0.
I have found that other DEs let set the number of steps of backlight and the change of the backlight is usually done by keyboard.
LXQt should add those shortcuts to lxqt-panel brightness plugin. 
Then lxqt-config-brightness should let you set the number of steps of backlight, the gamma (for desktop computers) and backlight. I think that an option for launch and config [gammastep](https://gitlab.com/chinstrap/gammastep) should be added also.

Thanks for your hard work.

tsujan:
> LXQt should add those shortcuts to lxqt-panel brightness plugin.

Good idea.

Issue #903, opened 2022-12-22
Missing kscreen includes cause build failures.


Chiitoo:
##### Expected Behavior

(I'd rather make a pull request but I haven't had the time to prepare it right... so I'll create this issue at least while I can!)

I'm not sure this is in any release version of KScreen yet, but there are some new requirements for headers (possibly related to 38113506382 [1]).

1. https://invent.kde.org/plasma/kscreen/-/commit/38113506382d28f7cb44a8ecd9911817649e6cde

##### Current Behavior

One example of a point of failure:

``` gcc
[...]/lxqt-config-monitor/monitorwidget.cpp:111:25: error: invalid use of incomplete type ‘class KScreen::Mode’
  111 |                 if( mode->size() == output->currentMode()->size() )
      |
```

##### Possible Solution

Add `#include <kscreen/mode.h>` (and friends) in `monitorwidget.cpp` and friends.

##### Steps to Reproduce (for bugs)

1. Build 'lxqt-config' with 'kscreen' from 'git master'.

##### Context

Unable to build 'lxqt-config' without patching it.

##### System Information

* Distribution & Version: Gentoo Linux
* Kernel: 6.0.9
* Qt Version: 5.15.7 (with KDE patches)
* liblxqt Version: git master
* lxqt-build-tools Version: git master
* Package version: git master

Thank you!

tsujan:
> ... with 'kscreen' from 'git master'.

A wild guess: they may be preparing kscreen for Qt6. If that's the case, the required changes will be in the Qt6 port of lxqt-config.

rikmills:
In a Ubuntu build against the Plasma 5.27 released beta tars:

```
In file included from /usr/include/KF5/KScreen/KScreen/Config:1,
                 from /<<PKGBUILDDIR>>/obj-x86_64-linux-gnu/lxqt-config-monitor/lxqt-config-monitor_autogen/EWIEGA46WW/../../../../lxqt-config-monitor/fastmenu.h:28,
                 from /<<PKGBUILDDIR>>/obj-x86_64-linux-gnu/lxqt-config-monitor/lxqt-config-monitor_autogen/EWIEGA46WW/moc_fastmenu.cpp:10,
                 from /<<PKGBUILDDIR>>/obj-x86_64-linux-gnu/lxqt-config-monitor/lxqt-config-monitor_autogen/mocs_compilation.cpp:2:
[...]

PR #915, opened 2023-02-02
lxqt-config-monitor: add more header file inclusion for libkscreen 5.26.90

mtasaka:
With https://github.com/KDE/libkscreen/commit/94f330959b0eda775418aef7faee80ce69144e63 , `#include <KScreen/Output>` no longer includes "mode.h" implicitly. So in lxqt-config-monitor, files using `class KScreen::Mode` should include `#include <KScreen/Mode>` explicitly.

Related: #903 .


palinek (approved):
GTM

Issue #937, opened 2023-05-03
Also use `~/.Xresources` to set cursor theme instead of or in addition to `~/.Xdefaults`


tsujan:
Because the default cursor theme may be set in `Xresources` inside `/etc`, in which case the default set by LXQt won't persist in next sessions.

See this comment by @humpity: https://github.com/lxqt/lxqt/issues/2391#issuecomment-1533247900

PR #938, opened 2023-05-03
Put cursor settings in `"~/.Xresources"`

tsujan:
`~/.Xdefaults` is the old place, altough it is also written to.

This patch fixes the problem of not being able to set the cursor (size) with `lxqt-config-appearance` when Xresources from `/etc` is read.

Credit goes to @humpity for finding the main cause.

Fixes https://github.com/lxqt/lxqt-config/issues/937

### DIFF
--- a/liblxqt-config-cursor/cfgfile.cpp
+++ b/liblxqt-config-cursor/cfgfile.cpp
@@ -54,23 +54,22 @@ QMultiMap<QString, QString> loadCfgFile(const QString &fname, bool forceLoCase)
     return res;
 }
 
-void fixXDefaults(const QString &themeName, int cursorSize)
+static inline void setXcursorInFile(const QString &fileName, const QString &themeName, int cursorSize)
 {
     QStringList lst;
     {
-        QFile fl(QDir::home().path()+QStringLiteral("/.Xdefaults"));
+        QFile fl(fileName);
         if (fl.open(QIODevice::ReadOnly))
         {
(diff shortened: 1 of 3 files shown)


tsujan:
To see the problem, edit `~/.Xresources` (create it if not existing) and put a big cursor size in it, like this:

```
Xcursor.size: 40
```
Then log out and log in. After that, you couldn't restore the cursor size with `lxqt-config-appearance`. But if you apply the patch, you could.

NOTE: The cursor size on the title-bars of KWin won't change until the next login. That's not related to `lxqt-config-appearance` but to KWin.

tsujan:
BTW, I didn't know from whom I should request a review — apparently, the code wasn't written by an LXQt dev.

tsujan:
Rebased (merge conflict fixed).

I wanted to make a point release with this fix because, IMO, the bug was serious. But it isn't possible anymore, after the changes related to `lxqt-menu-data` — it should have been merged before them :(

stefonarch:
That's a pity...  point releases for all components involved is probably an overkill.

tsujan:
>  point releases for all components involved is probably an overkill.

Yes, point releases should only be about critical bugs.

luis-pereira (approved):
Tested it and worked Ok.
Code also seems Ok.
GTM.

tsujan:
@luis-pereira, thanks for the review!

PR #944, opened 2023-05-24
Update the windowTitle string in applicationchooser.ui

### LINKED ISSUE #860: "ApplicationChooser" is not a good name
In File Associations → Associations, if you select a mimetype from the left tree and press "Change..." or "Choose..." in the right pane, the "ApplicationChooser" dialog will be opened. That name may be good for C++ coding but not for appearing inside a GUI.

Please change it to "Application Chooser" — just add a space.

isf63:
Change windowTitle string to include a space: "ApplicationChooser" → "Application Chooser".

Closes #860.

### DIFF
--- a/lxqt-config-file-associations/applicationchooser.ui
+++ b/lxqt-config-file-associations/applicationchooser.ui
@@ -11,7 +11,7 @@
    </rect>
   </property>
   <property name="windowTitle">
-   <string>ApplicationChooser</string>
+   <string>Application Chooser</string>
   </property>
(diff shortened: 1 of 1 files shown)


stefonarch:
This will undo all translations if we run `lxqt-transupdate` after. Probably just a search&replace for all ` lxqt-config-file-associations*.ts` "ApplicationChooser → Application Chooser" is  better.

tsujan (approved):
GTM.

> This will undo all translations 

@stefonarch, there's no other way; "ApplicationChooser" was a mistake.

Issue #925, opened 2023-03-11
[lxqt-config-monitor] primary monitor setting doesn't work properly


Loader009:
##### Expected Behavior

Primary monitor should be set - monitor number should not change.
##### Current Behavior

Primary monitor seems to always be the last monitor - the conf file shows - all monitors are "primary".
##### Steps to Reproduce (for bugs)

1. open lxqt-config-monitor
2. set another primary monitor and save/apply
3. close lxqt-config-monitor
4. open it again and check again

##### Context

After some update in the last few months I had to rearrange my panels due to wrong monitor number/order.
Some fullscreen applications only open on the primary display, which is currently not the one I actually use as my main display (a total of three displays connected).
##### System Information

* Distribution & Version: archlinux
* Kernel: 6.2.2-zen2-1-zen
* Qt Version: 5.15.8
* liblxqt Version: 1.2.0-1
* Package version: 1.2.0-1
--- I don't use AUR for lxqt anymore! ---

tsujan:
OK, the bug is in `libkscreen` 5.27.2.

The methods that worked fine were in `libkscreen` 5.26.5:

```c++
bool Output::isPrimary() const
{
    return d->primary;
}

void Output::setPrimary(bool primary)
{
    if (d->primary == primary) {
        return;
    }

    d->primary = primary;

    Q_EMIT isPrimaryChanged();
}
```

Now, these are their corresponding methods in `libkscreen` 5.27.2:

```c++
bool Output::isPrimary() const
{
    return d->enabled && (d->priority == 1);
}

void Output::setPrimary(bool primary)
{
    setPriority(1);
}
```

`setPrimary()` ignores its argument. My guess is that it's an incomplete draft that has popped up in the release by mistake (they have a plan to give priority to outputs, instead of making one of them primary).

tsujan:
The problems still exists in git.

I think we need to change the code based on output priorities, but it may be too soon for that.

Loader009:
I actually don't think that they have it on the screen -> https://bugs.kde.org/buglist.cgi?component=libkscreen&product=KScreen&resolution=---
I might try to downgrade the lib and lxqt-config instead.

Loader009:
Getting an error when I run lxqt-config-monitor (even after recompilation with libkscreen 5.26.5):
```
org.kde.kscreen: Invalid key in Output map:  "priority"
Connecting to KScreen...
Error: Config is invalid, probably backend couldn't load
```

[...]

PR #954, opened 2023-09-03
Fixed primary monitor after a change in `libkscreen`

tsujan:
By replacing `Output::setPrimary()` with `Output::setPriority()`.

Fixes https://github.com/lxqt/lxqt-config/issues/925


ratijas:
Ideally `output->setPrimary` should be replaced with `config->setPrimaryOutput`, or `config->setOutputPriority` for more granular approach. In fact, I was against `Output::setPriority` being a public method, but it was needed for config deserializers on backends.

PR #957, opened 2023-09-29
lxqt-config-locale: Add Save button and remove prompt dialog

isf63:
The dialog buttons are now Reset/Save/Close.

Also perhaps for the button name, Apply is a better as opposed to Save ?

Related to https://github.com/lxqt/lxqt/issues/2226.


isf63:
Necessary changes:
- The Close button should be set default (I didn't find anything for that in https://github.com/lxqt/liblxqt/blob/master/configdialog/lxqtconfigdialog.cpp)
- Save button should be disabled in the events which contain `hasChanged = true` in [1] `LocaleConfig::load` [2] `LocaleConfig::connectCombo`

isf63:
This is done, and it works, but it's a little hacky — specifically with use of `extern`.

PR #959, opened 2023-10-20
Don't return uninitialized struct member

luis-pereira:
At Result_moveMonitorPictureToNearest compareTwoMonitors() with:
    if(monitorPicture1Rect.intersects(monitorPicture2Rect)) {
        result.ok = true;
        return result;
    }
result.outside is returned without initialization.


tsujan (approved):
GTM

Issue #577, opened 2019-12-14
[lxqt-config-monitor] Two "Apply" buttons in the same screen


papoteur-mga:
After opening lxqt-config-monitor, then selected "Paramètres" (Settings ) I get a list of saved configurations, and under the box, a first button "Appliquer" (Apply) and a second button "Appliquer" at the bottom right. Two buttons with the same name are confusing.
1. After selecting a config to restore, the second button has no effect. There is no need of a button which has no effect.
2. After a click on the second button, new settings are indeed applied, but the application is directly closed. In other parts of the application,like applying quick settings, we have a dialogue to confirm that the configuration is OK, and after that the application survives. I'm not in favour of to much dialogues, but I don't understand why this is dealt so differently.
Release Mageia lxqt-config-0.14.1-2.mga7

the-loudspeaker:
![2020-01-21-162750_1366x768_scrot](https://user-images.githubusercontent.com/37928641/72799637-b1c70980-3c6b-11ea-834c-5cedf706ee6f.png)

I think the op refers to the "Save" and "Apply" buttons. The thing is, they are for different purposes. "Apply" button only applies the setting to current session. The "Save" button makes it consistent across all sessions. Checkout the Settings dialoguebox to understand better. It is updated after you press the save button. The settings by "Apply" button do not remain after a reboot. 

I think it is good like this. acts as a failsafe.

papoteur-mga:
Hello,
I don't speak about this window, but this one, got when pressing on "Settings".
![image](https://user-images.githubusercontent.com/5154580/72801359-9fd46f00-3c49-11ea-8b99-0fac9713ea52.png)
You can that there is 2 "Apply" buttons.

the-loudspeaker:
Oh that! Sorry my bad. Will have to look in more.

vc-01:
I checked the [source](https://github.com/lxqt/lxqt-config/blob/c6a164b35a25c4e9b2600f7a1d36050f2a62c82d/lxqt-config-monitor/settingsdialog.cpp#L26) and indeed, the "Apply" button isn't connected to any action.

Neither, I see any default action for Apply button in _LXQt::ConfigDialog_ ([source](https://github.com/lxqt/liblxqt/blob/a0fafe2228b87271a9b9c6491cc3dfdd494ca752/configdialog/lxqtconfigdialog.cpp#L70)).

It feels good to click it, but it does nothing :)

I see the following options to fix the UI:

[...]

PR #961, opened 2023-10-30
lxqt-config-monitor: Unintrusive (hopefully) UI improvements proposal

vc-01:
Greenlit by project team member.

Please see individual commit messages for detailed explanations for each change.

Refs: gh-577

### DIFF
--- a/lxqt-config-monitor/managesavedsettings.cpp
+++ b/lxqt-config-monitor/managesavedsettings.cpp
@@ -23,6 +23,7 @@
 #include "monitor.h"
 #include <QDebug>
 #include <QInputDialog>
+#include <QMessageBox>
 #include <QDateTime>
 
 Q_DECLARE_METATYPE(MonitorSavedSettings)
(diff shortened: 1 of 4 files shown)


tsujan:
Thanks!

Since you were worried about your PR, I'm here to emphasize that it'll be reviewed (if no other dev volunteers, I'll do it). But since LXQt 1.4.0 will be released in a week or so, and we won't have enough time for testing until then, reviewing/merging will be done after the new release.

vc-01:
Thank you for your very fast response. I'm more than fine if the review takes, say a full year of time for example, after all this is an unpaid project and I'm happy that it exists at all. I was worried more about going against ideas of the project and waste time working on something to be later rejected or frowned upon. Which is especially a case for no-one-need-it-no-one-want-it UI changes. And as mentioned already, the UI for lxqt-config-monitor needs much more than what's been proposed in this PR.

_-- No rush. --_

lxqt-config-monitor/managesavedsettings.cpp
```diff
@@ -164,6 +167,14 @@ void ManageSavedSettings::onApplyItem()
     if (ui.allConfigs->currentItem() == nullptr)
         return;
     MonitorSavedSettings settings = ui.allConfigs->currentItem()->data(Qt::UserRole).value<MonitorSavedSettings>();
+
+    if (not isHardwareCompatible(settings)) {
```
tsujan, line 171: Please change this to `if (!isHardwareCompatible(settings)) {`. In LXQt we don't use `and`, `or` or `not`.

vc-01: Sure.

tsujan:
Thanks! IMO, this is much better than the current master branch. There are other problems in the config dialog too, but this PR does its job well. So, if other members find no issue in it either, it could be merged.

Before I give it a "GMT", would you please squash all commits?

vc-01:
> ... would you please squash all commits?

I have squashed the commits as requested.

tsujan:
Assuming that there's no objection, I'm merging it — its goals are reasonable, and it works as intended.

PR #965, opened 2023-11-05
Make Locale Config comboboxes not editable

isf63:
Given that the comboboxes are formatted as: "\<Country\> - \<Language\> (\<LANG\>)", it's unlikely that someone would type their locale in manually. This simplifies the GUI; however, I have concerns in the case of when a locale is missing from the list somehow.


tsujan:
> it's unlikely that someone would type their locale in manually.

It isn't just for typing the locale, but rather for finding it. It has a completer popup.

isf63:
Ah. I never noticed that feature. Now it makes sense why it is editable and has a clear button.

PR #982, opened 2024-02-25
Expand initializer lists and use `default` keyword for ctors/dtors

isf63:


isf63:
I'm closing this as I want to revisit the specifics for initializer lists.

PR #986, opened 2024-03-19
Qt6 port

tsujan:


tsujan:
@stefonarch
I should reread the patch twice, but would you test it thoroughly under X11? I know how difficult that could be.

Please make sure that you don't have any parallel installation in `/usr/local/` before testing it!

PR #990, opened 2024-04-15
Locale config: Add 'Apply' button

isf63:
Replaces the confirmation message box upon closing with an Apply button.

My prior attempt (#957) used `extern`; this instead uses signals/slots.

### DIFF
--- a/lxqt-config-locale/localeconfig.cpp
+++ b/lxqt-config-locale/localeconfig.cpp
@@ -111,6 +111,7 @@ void LocaleConfig::load()
     connect(m_ui->checkDetailed, &QGroupBox::toggled, [ = ]()
     {
         updateExample();
+        emit enableApply(true);
         hasChanged = true;
     });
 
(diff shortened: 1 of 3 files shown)


isf63:
@tsujan, would you mind looking at this? It regards https://github.com/lxqt/lxqt/issues/2226

tsujan:
Oh, sorry, I thought my opinion was clear. I told it at https://github.com/lxqt/lxqt/issues/2226#issuecomment-1100784428; that's why I didn't review this PR, hoping that another dev might do it. I explain it again:

One of the advantages of an Apply button is that it may replace a prompt dialog. But in this case, the info on the prompt dialog is needed. Adding an Apply button and showing a prompt dialog defeats the purpose, IMHO.

Apart from this, I know that some LXQt devs, who were here before I joined LXQt, prefer the lack of an Apply button, So, I think we have a good balance now and don't need to force Apply buttons on more dialogs.

isf63:
The flow is changed from:
- Modified → Close → "Do you want to save your changes? They will take effect the next time you log in." (Yes / No)

with

- Modified → Apply button → "Settings will take effect the next time you log in."

For me the second is clearer, but it is a matter of taste. Some users/devs will prefer one to the other.

PR #992, opened 2024-04-23
lxqt-config-input: Fixes a FTBFS with Clang

luis-pereira:
It fails to build with, at least, clang version 16.0.6.

### DIFF
--- a/lxqt-config-input/touchpaddevice.cpp
+++ b/lxqt-config-input/touchpaddevice.cpp
@@ -368,7 +368,12 @@ bool TouchpadDevice::setTapToDragEnabled(bool enabled) const
 
 bool TouchpadDevice::setAccelSpeed(float speed) const
 {
-    return set_xi2_property(LIBINPUT_PROP_ACCEL, QList<QVariant>({speed}));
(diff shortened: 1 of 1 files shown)


tsujan:
@luis-pereira 
Thanks! And sorry for the delay; will check your recent PRs tomorrow. We may also need point releases.

luis-pereira:
> Thanks! And sorry for the delay; will check your recent PRs tomorrow. We may also need point releases.

This one is annoying to me. Clang is my default compiler.

tsujan:
What I don't get is that `QVariant::QVariant(float val)` [exists](https://doc.qt.io/qt-6/qvariant.html#QVariant-11). Why should Clang result in FTBFS?

luis-pereira:
@tsujan `QVariant` is not the issue here.
Clang is very fussy with narrowing in initializer list. That's the issue.
Updated the PR to be more easily understandable, I hope.

tsujan:
@luis-pereira 
Yes, I saw you code comment. IMO, this is much better.

tsujan (approved):
GTM

PR #994, opened 2024-05-09
Appearance save size use after free

luis-pereira:
Also found by ASAN.


tsujan:
I agree that the original code seems problematic with `Qt::WA_DeleteOnClose`, but, for whatever reason (which I haven't investigated), it correctly saves the size when the dialog is closed *in any way*, e.g., by the WM or by pressing `Escape`. That's while the patch saves the size only on pressing the Close button.

tsujan (approved):
Yes, this may be the best way.

Issue #916, opened 2023-02-02
lxqt-config-mon crashes on wayland except kwin_wayland


stefonarch:
No suprise:
```
$ lxqt-config-monitor 
kscreen.kwayland: Connection to Wayland server timed out.
Connecting to KScreen...
Error: Config is invalid, probably backend couldn't load

```
At every wakeup from suspend `journalctl -r` shows its coredump. 
```

Core was generated by `/usr/bin/lxqt-config-monitor -l'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fc11bca0868 in KScreen::Config::outputs() const () from /usr/lib/libKF5Screen.so.7
[Current thread is 1 (Thread 0x7fc119402840 (LWP 10822))]
e(gdb) where
#0  0x00007fc11bca0868 in KScreen::Config::outputs() const () at /usr/lib/libKF5Screen.so.7
#1  0x000055e1dc48eecf in applySettings(QSharedPointer<KScreen::Config>, QList<MonitorSettings>)
    (config=..., monitors=...) at /home/stef/git/lxqt/lxqt-config/lxqt-config-monitor/loadsettings.cpp:114
#2  0x000055e1dc48e89f in operator()(KScreen::ConfigOperation*) const
    (__closure=0x55e1dd2a3f40, op=0x7fc1100143b0)
    at /home/stef/git/lxqt/lxqt-config/lxqt-config-monitor/loadsettings.cpp:71
#3  0x000055e1dc48f924 in QtPrivate::FunctorCall<QtPrivate::IndexesList<0>, QtPrivate::List<KScreen::ConfigOperation*>, void, LoadSettings::applyBestSettings()::<lambda(KScreen::ConfigOperation*)> >::call(struct {...} &, void **) (f=..., arg=0x7ffe03535900) at /usr/include/qt/QtCore/qobjectdefs_impl.h:146                                    
#4  0x000055e1dc48f8e5 in QtPrivate::Functor<LoadSettings::applyBestSettings()::<lambda(KScreen::ConfigOperation*)>, 1>::call<QtPrivate::List<KScreen::ConfigOperation*>, void>(struct {...} &, void *, void **)                    
    (f=..., arg=0x7ffe03535900) at /usr/include/qt/QtCore/qobjectdefs_impl.h:256
[...]

PR #995, opened 2024-05-15
Show message when under wlroots instead of exiting with error

stefonarch:
Related to https://github.com/lxqt/lxqt-config/issues/916


tsujan (approved):
GTM

PR #997, opened 2024-05-17
Show unsupported message under wayland

stefonarch:
Maybe we could change this in futere to a dialog "Open compositor settings if available?"
* if kwin:
A list to open keyboard settings `kcmshell6 kcm_keyboard`  or just launch `systemsettings`

* if wayfire launch `wcm`
* if labwc launch `labwc-tweaks` (atm not complete)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Configures hardware of pointer devices. Settings like repeat delay and interval
 of keyboards or acceleration of pointer devices.
 
 Binary `lxqt-config-input`.
+**Note**: Not supported currently under wayland.
 
 #### Locale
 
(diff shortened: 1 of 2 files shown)


stefonarch:
This has a bad side effect: at every login the message is shown, as `lxqt-config-input --load-touchpad` is executed.

stefonarch:
Maybe we should disable loading touchpad settings under wayland. It does just nothing, no harm too.
https://github.com/lxqt/lxqt-session/blob/fb537d535742f216c9bd113a575d163abf3cacae/lxqt-session/src/sessionapplication.cpp#L105

tsujan (approved):
GTM

Issue #1004, opened 2024-06-14
lxqt-config-monitor crashing with kscreen on openSUSE Tumbleweed


bbigam:
lxqt-config-monitor crashes on every login, every wake from sleep, and every user launch of the settings module. It segfaults and coredumps, with kscreen.

##### Expected Behavior
l-c-monitor doesn't crash and coredump in any standard usage. If kscreen crashes, l-c-monitor errors rather than crashing and coredumping.

##### Current Behavior
I don't have any particular need for l-c-monitor, since I use only one monitor, my laptop display.  But opening it shows a window that's blank as if there's not detected display, and when I close it, it and kscreen segfault and coredump. If I check journalctl, it shows similar segfaults and coredumps for l-c-monitor on my user session login or wake from sleep, and the cordumps pile up in `/var/lib/systemd/coredump/`
Since there's not user-facing option to disable l-c-monitor, the coredumps just keep piling up.

Example from journalctl in wake from sleep:
```
Jun 14 07:11:33 localhost.localdomain lxqt-config-monitor[20728]: [load.applyBestSettings()] Finished
Jun 14 07:11:33 localhost.localdomain lxqt-config-monitor[20728]: lxqt-config-monitor: Applying current settings...
Jun 14 07:11:33 localhost.localdomain kernel: lxqt-config-mon[20728]: segfault at 10 ip 00007fedb9430930 sp 00007ffc2ed1a208 error 4 in libKF6Screen.so.6.0.5[7fedb942a000+2d000] likely on CPU 1 (core 1, socket 0)
Jun 14 07:11:33 localhost.localdomain systemd-coredump[20756]: Process 20728 (lxqt-config-mon) of user 1000 dumped core.
                                                               #1  0x000056150392127b n/a (lxqt-config-monitor + 0x2427b)
                                                               #2  0x0000561503923135 n/a (lxqt-config-monitor + 0x26135)
                                                               #16 0x000056150390ec10 n/a (lxqt-config-monitor + 0x11c10)
                                                               #19 0x000056150390ec85 n/a (lxqt-config-monitor + 0x11c85)
```

##### Possible Solution
I have no idea how to fix the crash, and don't know if it's an lxqt or downstream issue. Having a GUI option to disable l-x-monitor would mitigate the issue for those who don't need it anyway.

##### Steps to Reproduce (for bugs)
1. Log in to an LXQt session, or wake from sleep within one (not just from screen lock)
2. OR open l-c-monitor
[...]

PR #1005, opened 2024-06-14
Added nullity check for monitor config

tsujan:
*Supposedly* fixes https://github.com/lxqt/lxqt-config/issues/1004


tsujan:
Since this is nothing more than a simple nullity check, and because it prevents the crash of the linked report, I'm merging it. Hopefully it might also prevent the problem reported at https://github.com/lxqt/lxqt/issues/2509.

PR #1009, opened 2024-07-08
Added hint for cursor settings under wayland

stefonarch:


stefonarch:
I see, but we have to consider also titlebars from the compositor, so better setting it there, if I'm not mistaken.

stefonarch:
Made some tests and it looks like at least under labwc if no theme and size are set it uses LXQt setting for the decorations (did not set GTK theme so gtk windows had the previuos cursor but the Qt cursor on the SSD decoration. 
Will test the the other WM after.

PR #1012, opened 2024-07-23
Do not load monitor or input settings under Wayland

tsujan:
Because, otherwise, crash will happen.

KWin is automatically excluded from this for monitor settings.

### DIFF
--- a/lxqt-config-input/lxqt-config-input.cpp
+++ b/lxqt-config-input/lxqt-config-input.cpp
@@ -35,6 +35,8 @@
 int main(int argc, char** argv) {
     LXQt::SingleApplication app(argc, argv);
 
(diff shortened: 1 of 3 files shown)


marcusbritanicus:
FYI: WayQt has support for output management: resolution, refresh rate, scaling, and so on. So it's possible to provide some support for wlroots based compositors that support this protocol.

snapshot-2024-07-24-10:02:22

stefonarch:
>for wlroots based compositors that support this protocol.

Basically for those where `kanshi` and wl-displays` work I guess.

stefonarch (approved):
`...you could try kanshi to configure your monitor(s)`

Include `wdisplays` here?

Works fine everywhere.

tsujan:
Here we just mentioned a well-known tool in a short message. More tools are appearing as Wayland becomes more user-friendly. Of course, Qt-based solutions like WayQt are preferable with LXQt.

Issue #1013, opened 2024-07-26
Show "GenericName" instead of "Name" in lxqt-config in tooltips


stefonarch:
##### Expected Behavior

As in the menu:

![immagine](https://github.com/user-attachments/assets/9a3dd01f-3359-4c75-8ba6-af4b6cf7a616)

##### Current Behavior

Tooltip is useless for more info
![immagine](https://github.com/user-attachments/assets/b40673e2-7737-49cc-a267-e4191c1a2d6b)
#### Context

This is "Schermi" (Displays) fopr wdisplay, but our "Monitor" isn't more explicative too.
* Distribution & Version: any

Issue #494, opened 2019-06-15
Tooltips aren't informative


tsujan:
##### Expected Behavior

Tooltips are usually used to add something that doesn't exist in the GUI. A desktop file can contain a "Comment" section and the main menu (of lxqt-panel) shows it. Why not showing it here?
##### Current Behavior

In the case of lxqt-config, tooltips only show full names. OK, that's good because the GUI names may be elided but it isn't enough.
##### Context

My recent obsession with tooltips in libfm-qt (done).
##### System Information

Latest git

tsujan:
BTW and just as a reminder, lxqt-panel's main menu apparently prefers GenericName to Comment for the tooltip; it shows Comment only when there's no GenericName . That's far from ideal, IMO.

agaida:
@tsujan - your are absolutely right, no discussion about :)

tsujan:
Good to know that it isn't a problem only from *my* perspective :)

PR #1014, opened 2024-07-26
More informative tooltips in lxqt-config dialog

tsujan:
Closes https://github.com/lxqt/lxqt-config/issues/494 and closes https://github.com/lxqt/lxqt-config/issues/1013

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -152,15 +152,18 @@ public:
             auto name = m_list[index.row()].xdg().name();
             auto gName = m_list[index.row()].xdg().localizedValue(QStringLiteral("GenericName")).toString();
             if (!gName.isEmpty() && QString::compare(gName, name, Qt::CaseInsensitive) != 0)
-                toolTip = name + QStringLiteral("\n") + gName;
+            {
+                toolTip = QStringLiteral("<p><b>") + name
+                          + QStringLiteral("</b></p><p><i>") + gName + QStringLiteral("</i></p>");
+            }
             else
-                toolTip = name;
+                toolTip = QStringLiteral("<p><b>") + name + QStringLiteral("</b></p>");
             auto comment = m_list[index.row()].xdg().comment();
             if (!comment.isEmpty()
                 && QString::compare(comment, name, Qt::CaseInsensitive) != 0
                 && QString::compare(comment, gName, Qt::CaseInsensitive) != 0)
             {
-                toolTip += QStringLiteral("\n") + comment;
+                toolTip += QStringLiteral("<p>") + comment + QStringLiteral("</p>");
             }
             return toolTip;
         }
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -144,8 +144,26 @@ public:
 
     QVariant data(const QModelIndex &index, int role) const override
     {
-        if (role == Qt::DisplayRole || role == Qt::ToolTipRole)
+        if (role == Qt::DisplayRole)
             return m_list[index.row()].xdg().name();
+        if (role == Qt::ToolTipRole)
+        {
+            QString toolTip;
+            auto name = m_list[index.row()].xdg().name();
+            auto gName = m_list[index.row()].xdg().localizedValue(QStringLiteral("GenericName")).toString();
+            if (!gName.isEmpty() && QString::compare(gName, name, Qt::CaseInsensitive) != 0)
+                toolTip = name + QStringLiteral("\n") + gName;
+            else
+                toolTip = name;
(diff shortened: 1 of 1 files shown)


stefonarch (changes requested):
immagine
It's too much now, at least _Name_ has not to be replicated.

tsujan:
> It's too much now

It isn't. Use `LANG=de lxqt-config` and you'll see why:

german

stefonarch:
This is an example in english and it's enough, most are like that - terrible to read as they are redundant, those without comments like `systemsettings` are slightly better.  
If that is not too much I don't know:
immagine

immagine

Just showing "GenericName" or "Comment" will do.

tsujan:
> This is an example in english

English isn't the only Language that we support, and I had items in English whose names were elided. The whole name is guaranteed to be shown in the tooltiip.

Also, see my old report: "OK, that's good because the GUI names may be elided but it isn't enough."

tsujan:
> terrible to read as they are redundant

That depends on how the desktop file is written. For example, nothing is redundant for "Keyboard and Mouse", And not every desktop file is like LXQt ones.

tsujan:
Why did you edited *my* comments?!  :D

tsujan:
> Sorry but I'll never approve this as it is.

That's unfortunate, because I insist on its current form.

The code guards against identical strings, but no code can detect poorly written desktop files. Someone might have been hasty when writing some LXQt desktop entries ;)

stefonarch:
> Why did you edit _my_ comments?! :D

Oops. Was tired and hit "Edit" instead of "Quote reply", sorry.

>The code guards against identical strings

The name is identical, and in most desktop files they don't know really what to add in "Comment" and write _mostly_ the same as in "GenericName". If the code would check for "identical meaning" but it can't. 

>>    I never saw any elided text

>Is that really relevant? I intentionally attached a screenshot with German.

Only because German is the only language which creates those long attachments of substantives we shouldn't ruin the tooltips for all others. If we include the name only if it's more than 13 characters? 

And remove "Comment", reason see above.

 We could eventually fix and unify all _our_ `.desktop` files, only 3 or 4 have informative comments atm. Some are even wrongly translated I see, like https://github.com/lxqt/pcmanfm-qt/blob/master/pcmanfm/translations/pcmanfm-qt-desktop-pref.desktop.yaml in italian is "Configure desktop path and appearance", the german one talks about ""Settings such as background image or menus used", and who knows in other languages what can be seen now, portughese has an issue too:
```
Desktop Entry/Name: "Ambiente de trabalho"
Desktop Entry/GenericName: "Definições do ambiente de trabalho"
Desktop Entry/Comment: "Mudar os papéis de parede e o comportamento do gestor do ambiente\
  \ de trabalho"
  ```

Compromise:  
* Show only names if longer than 13 chars or what the limit is for eliding
* Fix all our GenericNames and Comments.

**Comment by tsujan** [comment]
IMO, this simple change doesn't deserve more discussion, and I don't understand your reaction. pcmanfm-qt's item tooltips also include file names and no one has found them "redundant" or "terrible".

The PR reveals that generic names were forced on some LXQt desktop entries without thinking whether they were needed. If there's something "redundant", it's that. An app called FeatherPad can have a generic name, but the name of an app called "Monitor Settings" is already generic.

**Comment by stefonarch** [comment]
In pcmanfm-qt the filename is bold and often the tooltip covers the item. 

My point from the beginning  is that tooltips are per _additional_ information (which _could_ include omitted elided text for monsterwords like the german ones but   `only for them). 
Every german or other user will recognise easily the elided word also without tooltip and I repeat all that  for the last time too :)

My averse reaction at first sigfht  was  because most `.desktop` files are  really redundant in their texts, this can be fixed for ours. Nowhere ever "GeneralName" and "Comments" were exposed together, some include "LXQt", some not, IMO "LXQt" should be removed from them as they are shown only under LXQt. 

```
Desktop Entry/Name: "Brightness"
Desktop Entry/GenericName: "Brightness Settings"
Desktop Entry/Comment: "Configure brightness"

```
could be:
```

Desktop Entry/Name: "Brightness"
Desktop Entry/GenericName: "Brightness Settings"
Desktop Entry/Comment: "Adjust backlight and brightness"
```

tsujan:
> In pcmanfm-qt the filename is bold

It's bold here too.

> My averse reaction at first sight was because most .desktop files are really redundant in their texts

Not a problem of this PR.

As for how `GeneralName` should be written, I added a hint in my previous comments (although the current generic names don't look terrible to me): If `Name` is generic, either remove `GeneralName` or make it identical to it. Anyway, _that isn't related to the PR._ If it needs a discussion, here isn't the place.

stefonarch (approved):

PR #1015, opened 2024-07-27
Standardized desktop files for settings

stefonarch:
Draft for standardize configuration desktop files. At a first glance only globalkeys needs updating.  Related to https://github.com/lxqt/lxqt-config/pull/1014

### DIFF
--- a/lxqt-config-appearance/translations/lxqt-config-appearance.desktop.yaml
+++ b/lxqt-config-appearance/translations/lxqt-config-appearance.desktop.yaml
@@ -1,3 +1,3 @@
 Desktop Entry/Name: "Appearance"
 Desktop Entry/GenericName: "Appearance settings"
-Desktop Entry/Comment: "Appearance settings for LXQt"
+Desktop Entry/Comment: "Configure theme, colors, font, icons and more"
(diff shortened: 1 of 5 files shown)


tsujan:
The current changes are good. However,

 * Identical names and generic names (like in `lxqt-config-monitor`.desktop) don't cause a problem. KDE also has them in `systemsettings.desktop`.
 * Also, as far as I know, `GenericName` isn't a must when `Name` is generic — or even otherwise. Several desktop files (outside LXQt) lack it.

stefonarch:
>GenericName isn't a must when Name is generic 

We show `GenericName` in the panel menus, not sure what happens without them.

tsujan:
Where? I don't remember it. But if panel had a problem without it, it would be show up, because there are several desktop entries without `GenericName`.

stefonarch:
> Where? I don't remember it. But if panel had a problem without it, we should have seen it, because there are several desktop entries without `GenericName`.

In the tooltip. Made a test, if `GenericName` is empty `Comment` is shown.

tsujan:
Yes. So, there's no problem; its absence is taken into account.

isf63:
Also for Appearance/Locale/Input -- "settings" in GenericName should be capitalized.

tsujan (approved):
GTM

Issue #1022, opened 2024-08-10
XSETTINGS not applied correctly


sledgehammer999:
I have been evaluating LxQt in a VM for a possible future migration of my main pc to it.
I noticed a few pain points and started to investigate. Luckily the code is hosted on github and it's online viewer/browser does wonders into researching the codebase without needing to clone the repos locally.
I am also knowledgeable with Qt and C++ and I am willing to implement solutions if there is agreement to the direction.
Enough chit chat, let me write about what I noticed.

##### Expected Behavior

Settings from `lxqt-config-appearance` should be applied consistently for Qt and Gtk apps. Settings should restore on session start.
Available mechanisms for realtime "settings change" signaling to running apps should be used effectively.

##### Current Behavior

Some options from `lxqt-config-appearance` are applied inconsistently between Qt and Gtk. I can't remember all the stuff but I focused on the Font settings. All these settings are directly written to fontconfig files but they aren't saved into an lxqt config file.
Qt/Gtk apps seem to pick up the font family from fontconfig but not the DPI setting. They pickup the DPI settings if I shoehorn it into the config file of `xsettingsd`. Same for Firefox.
For whatever reason, writing the [gtk-xft-dpi](https://docs.gtk.org/gtk4/property.Settings.gtk-xft-dpi.html) setting into `settings.ini` doesn't work for Gtk apps. Did Gtk deprecate it without notice?

For Qt apps, they can pickup a lot of those settings from the `qplatformtheme` plugin. Even though [lxqt-qtplugin](https://github.com/lxqt/lxqt-qtplugin) has [code](https://github.com/lxqt/lxqt-qtplugin/blob/28bafb30a1f0f819d0d988ee5c980cafc1676f5f/src/lxqtplatformtheme.cpp#L408) to inform the Qt apps about the normal and fixed fonts, the variables are [initialized](https://github.com/lxqt/lxqt-qtplugin/blob/28bafb30a1f0f819d0d988ee5c980cafc1676f5f/src/lxqtplatformtheme.cpp#L239-L251) from non-existent config settings (`lxqt-config-appearance` writes the value only to fontconfig).

[...]

PR #1023, opened 2024-08-16
Write Fonts settings to LXQt config

sledgehammer999:
Partially implements #1022

With this, functionality is not lost or changed.
The fontsconfig is still written, however all related values are now saved/loaded to `lxqt.conf` first.

There's one caveat because I don't your policy on this. If this gets merged as-is the user's old settings will be lost. The user will need to customize them again.
Do you want a migration code that will load the customized settings into the lxqt.conf from fonts.conf if they are missing? It will makes things much more complicated though...

If this gets approve the future PRs will be in these stages:
* stage 2: Write the Fonts settings into the appropriate values of xsettings/gsettings/settings.ini
* stage 3: Yank out the code for xsettings and migrate it to lxqt-session. Make the xsettings deamon start at session start and continue running for the whole session. Reload it when changes are done to the settings.
* stage 4: Investigate what needs to be done on the lxqt-qtplugin side.

### DIFF
--- a/lxqt-config-appearance/configothertoolkits.cpp
+++ b/lxqt-config-appearance/configothertoolkits.cpp
@@ -41,6 +41,8 @@
 #include <sys/types.h>
 #include <signal.h>
 
+#include "fontconfigfile.h"
+
 static const char *GTK2_CONFIG = R"GTK2_CONFIG(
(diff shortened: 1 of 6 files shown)


sledgehammer999:
Don't get confused by the diff. The actual changes are minimal. I moved code around and that resulted in a big diff.

Issue #1030, opened 2024-09-03
[Feature request] Tooltips styling


stefonarch:
Some elements aren't styled according to palette colors, like tooltips in pcmanfm-qt.

tsujan:
Tooltips aren't touched anywhere because their colors aren't added to the palette (yet).

The styled tooltips you may see exist only in those components that are styled by LXQt Style, which uses Qt's style-sheet engine. There are are only a few of them: most importantly, Panel, Runner, and notifications. All other Qt apps are styled by the active Qt style (Fusion, Breeze, Kvantum,...).

isf63:
> styled by LXQt Style [...] Panel, Runner, and notifications

Is Configuration Center the only other one?

tsujan:
> Is Configuration Center the only other one?

Maybe. But it's better to use style-sheets with it only for its icon size and, perhaps, tooltips. Anything else can easily become problematic because of the clash between style-sheet and Qt style.

tsujan:
The new GUI will be like this:

![tooltips](https://github.com/user-attachments/assets/233cce48-65e1-4313-a0c8-11ff02c9a82c)

I tested it and saw no problem.

Issue #974, opened 2024-01-16
[Feature request] Add checkbox to apply corresponding palette to theme automatically


stefonarch:
We have some palette with the theme name, setting automatically those when switching theme would be a nice improvement as theming itself is still quite complicated, therefor adding a checkbox checked by default in `lxqt-config-appearance`.

If no corresponding name is present in `/usr/share/lxqt/palettes/` switch to default light palette.

##### Expected Behavior

It applies palette allover
##### Current Behavior

User has to explicitly set it.

PR #1031, opened 2024-09-03
Added tooltip colors to LXQt Appearance

tsujan:
Closes https://github.com/lxqt/lxqt-config/issues/1030


stefonarch (approved):
Works fine. 
Wondering if https://github.com/lxqt/lxqt-config/issues/974 would be hard to implement.
Our default palettes needs updates after merging.

tsujan:
> Wondering if https://github.com/lxqt/lxqt-config/issues/974 would be hard to implement.

I haven't checked it but don't think it's a piece of cake ;)

> Our default palettes needs updates after merging.

Old palettes work without problem — the change is backward compatible —  but you could add tooltip colors to them and update them.

Issue #978, opened 2024-02-09
Set "use wallpaper from theme" broken for local themes


stefonarch:
##### Expected Behavior

Wallpaper should change
##### Current Behavior

Wallpaper remains the same, or if unset always black screen with any theme/wallpaper combination.
##### Steps to Reproduce (for bugs)

1. Copy one or more themesfrom `/usr/share/lxqt/themes` to` ~/.local/share/lxqt/themes`
2.  Select "Use wallpaper from theme" and select one of those theme
3. no wallpaper change

##### System Information

* PcManFm-qt version:git, all<

EDIT: updated for reproducing

tsujan:
> PcManFm-qt version:git, wip_qt6

→ https://github.com/lxqt/lxqt-build-tools/pull/77#discussion_r1247857714

tsujan:
I wasn't able to reproduce this. I put a theme inside `~/.local/share/lxqt/themes/` and the wallpaper folder inside `~/.local/share/lxqt/`, and the wallpaper was changed when I selected and applied that (user) theme.

I also tested with a personal theme, whose wallpaper was inside its folder. Again, the wallpaper was changed correctly.

stefonarch:
Actually  I never looked at the `wallpaper.cfg` ... it's a relative path `../../wallpapers/` and I've no wallpapers there.

PR #1033, opened 2024-09-04
Added checkbox to use palette provided by LXQt theme

tsujan:
Closes https://github.com/lxqt/lxqt-config/issues/974

Also,
 * Save and restore the checked state of the checkbox for wallpaper overriding;
 * Swapped the rows of tooltip and link colors in the Qt Palette GUI.

### DIFF
--- a/lxqt-config-appearance/lxqtthemeconfig.cpp
+++ b/lxqt-config-appearance/lxqtthemeconfig.cpp
@@ -35,6 +35,7 @@
 #include <QMenu>
 #include <QDesktopServices>
 #include <QUrl>
+#include <QDir>
 
 #include <XdgDirs>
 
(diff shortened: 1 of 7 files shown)


tsujan:
@stefonarch 
For the first time in the history of mankind I didn't test my PR. Your thorough tests would be appreciated.

stefonarch (approved):
Works perfectly, initially wallpapers didn't update but I've local themes symlinks to git, so I ran into https://github.com/lxqt/lxqt-config/issues/978

I just added palettes for those missing to make changes with all themes. I think just adding later the same light palette for all light themes is the easiest way. Silver-dark needs renaming to "Silver".

tsujan:
Thanks for the tests!

> so I ran into https://github.com/lxqt/lxqt-config/issues/978

Will check it later and make a separate PR if any.

stefonarch:
No idea what happened, I was testing the palette PR on the testuser (and my user too) and found now "Use wallpaper from theme"  broken: it _always_ changes the wallpaper if checked or unchecked.

It it ok at your side?

tsujan:
> It it ok at your side?

Yes. I tried to reproduce your other report, and it worked fine.

stefonarch:
Compiled `lxqt-config` on the debian PC, same issue, wallpapers change always. Which other report?

tsujan:
> Which other report?

https://github.com/lxqt/lxqt-config/issues/978

tsujan:
I tested again right now The wallpaper is changed only when the box is checked, and the checked state is remembered correctly.

tsujan:
For reference:

@stefonarch and I found out why I couldn't see the issue he mentioned above, and also that the issue was reported 8 years ago! See https://github.com/lxqt/lxqt-config/pull/1036

isf63:
I'm getting @stefonarch's bug. Checked "use wallpaper" and applied Clearlooks, then unchecked and applied Arch-Colors, and Arch-Colors' wallpaper was applied. Is it because the checkbox state is only set by the Apply button?

PR #1036, opened 2024-09-05
Fixed an old problem in setting wallpapers

tsujan:
Closes https://github.com/lxqt/lxqt/issues/1186, a very old issue that surfaced recently.

### DIFF
--- a/lxqt-config-appearance/lxqtthemeconfig.cpp
+++ b/lxqt-config-appearance/lxqtthemeconfig.cpp
@@ -61,19 +61,6 @@ protected:
     }
 };
 
-/*!
- * \brief Check if currently configured wallpaper (read from pcmanfm-qt's
- * settings) is the same as \param themeWallpaper
- */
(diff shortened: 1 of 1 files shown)


stefonarch (approved):
This is fine now. While testing on the testuser I ran into that with "breeze" widget style:

invisible

After deleting anything beginning with "k" in `~/.config` it was gone. There was no way to get back visibility, even restarting the session. I tested a plasma session there some day ago.

What I didn't really pay attention was to icon themes, mostly it looks fine at a first glance, but dark/light icon sets could be changed also automatically. No idea if that is doable, palettes should have a marker for that I think.

tsujan:
> I ran into that with "breeze" widget style

Breeze takes some of its colors from KDE color scheme and overrides Qt palette. That may have become worse with Qt6, but I haven't checked.

> but dark/light icon sets could be changed also automatically

Checking "Colorize icons based on widget style" is the most that can be done.

PR #1037, opened 2024-09-06
Added an option for toolbar icon size

tsujan:
NOTE 1: If the toolbar of an app like pcmanfm-qt gets misaligned after changing the size, just close and reopen its window.

NOTE 2: This option is for Qt apps of course. Codes could be added later for GTK apps too, but I don't believe it's a good idea.

### DIFF
--- a/lxqt-config-appearance/styleconfig.cpp
+++ b/lxqt-config-appearance/styleconfig.cpp
@@ -54,6 +54,7 @@ StyleConfig::StyleConfig(LXQt::Settings* settings, QSettings* qtSettings, QWidge
 
     connect(ui->qtComboBox, QOverload<int>::of(&QComboBox::activated), this, &StyleConfig::settingsChanged);
     connect(ui->toolButtonStyle, QOverload<int>::of(&QComboBox::activated), this, &StyleConfig::settingsChanged);
+    connect(ui->toolBarIconSize, QOverload<int>::of(&QComboBox::activated), this, &StyleConfig::settingsChanged);
     connect(ui->singleClickActivate, &QAbstractButton::clicked, this, &StyleConfig::settingsChanged);
 
     connect(ui->winColorLabel, &ColorLabel::colorChanged, this, &StyleConfig::settingsChanged);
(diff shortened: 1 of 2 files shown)


stefonarch (commented):
Only with Breeze it looks fine here:

immagine

immagine

I can resize the window some more, but it's like that.

tsujan:
> Only with Breeze it looks fine here

What do you mean by "only with Breeze"? It works especially with Fusion, which our main target.

tsujan:
With Kvantum:

kvantum

With Fusion (launched with` lxqt-config-appearance -style fusion`):

fusion

That patch is simple and has nothing to do with the window size.

tsujan:
Oh... :)

If you mean the sizes of combo-boxes or the alignment of texts, not only it has nothing to do with the PR but is determined by the style (and its themes). For example, this is with Kvantum again, but with a different Kvantum theme:

kvantum

Strange that you hadn't noticed it before ;)

isf63:
It's probably my error, but I cannot get it to apply the icon size. I tried `sudo make install` and also `sudo cp ./lxqt-config-appearance/lxqt-config-appearance /usr/bin/` - the combobox is there but nothing is applied.

tsujan:
> It's probably my error, but I cannot get it to apply the icon size.

You need https://github.com/lxqt/lxqt-qtplugin/pull/93 (that's merged now). After upgradint to it and restarting apps, please use Fusion for testing (see the tooltip).

isf63:
Ok, that made it work. I use Fusion typically, its aesthetic doesn't bother me. 22px is an interesting change from 24px. :)

tsujan:
Thanks for testing!

> I use Fusion typically, its aesthetic doesn't bother me.

It's simple and clean.

> 22px is an interesting change from 24px. :)

Also, try 16px; you might like it if you get used to it.

stefonarch:
So there is no way to add an expandable spacer to cut down the length of the button?

tsujan:
That's a "form layout", and _it has always been so_. Breeze centers it, Fusion doesn't, and Kvantum has an option for centering it.

This image shows a form layout of qpdfview drawn by Fusion (left) and by a Kvantum theme that centers form layouts (right):

form_layout

stefonarch (approved):
GTM

PR #1038, opened 2024-09-06
Don't start `xsettingsd` outside X11

tsujan:
Also made sure that `ConfigOtherToolKits::setXSettingsConfig()` would be called only under X11.


stefonarch (approved):
GTM

PR #1042, opened 2024-09-09
Make sure Monitor Settings dialog is shown completely on Wayland

tsujan:
This is rather a simple workaround for undersized dialogs on Wayland (a new Qt bug), but it's good on X11 too.

### DIFF
--- a/lxqt-config-monitor/monitorsettingsdialog.cpp
+++ b/lxqt-config-monitor/monitorsettingsdialog.cpp
@@ -87,6 +87,9 @@ MonitorSettingsDialog::MonitorSettingsDialog() :
     });
 
     connect(ui.settingsButton, &QToolButton::clicked, this, &MonitorSettingsDialog::showSettingsDialog);
+
+    // make sure that widgets are shown completely (a Qt bug under Wayland?)
+    resize(sizeHint().expandedTo(QSize(600, 400)));
 }
 
 MonitorSettingsDialog::~MonitorSettingsDialog()


stefonarch (commented):
IMO the size could be a little bit bigger still.
Schermata_20240910_091209

tsujan:
The goal isn't setting the size but circumventing the Qt bug. A "good" size is different for different users and depends on locale, font, style,...

tsujan:
> it will not be remembered

That's a totally separate topic. I prefer to make a separate PR for it after this is merged.

stefonarch (approved):

stefonarch:
Ok. I think we should also change the message at login, will look into it. It's way too talkative IMO.

tsujan:
> It's way too talkative IMO.

This short message is verbose?

_LXQt monitor settings are currently unsupported under this wayland compositor._

_You can configure your monitor(s) with kanshi._

stefonarch:
No, this one:
_Default monitor settings have been applied. Use lxqt-config-monitor for adjusting your monitor settings._
 And you see it at every login with kwin and it makes saves notifications talking more then half the screen too.

tsujan:
> And you see it at every login

It's shown only when saved configs can't be applied, and that's a sign of problem. The message clearly tells that the user needs to change his saved settings or remove them to get rid of it.

If it's shown in another case, the code should be corrected.

tsujan:
I found no problem in the main code. However, it's better to change the notification to

```c++
                        mNotification->setSummary(tr("Default monitor settings applied."));
                        mNotification->setBody(tr("Use lxqt-config-monitor for adjusting your monitor settings."));
```

PR #1043, opened 2024-09-10
Remember size of Monitor Settings dialog

tsujan:
Also, split the notification to summary and body.

### DIFF
--- a/lxqt-config-monitor/loadsettings.cpp
+++ b/lxqt-config-monitor/loadsettings.cpp
@@ -95,7 +95,7 @@ void LoadSettings::applyBestSettings()
                         qDebug() << "lxqt-config-monitor: Extended mode has been applied";
 
                         mNotification->setSummary(tr("Default monitor settings applied."));
-                        mNotification->setBody(tr("Use lxqt-config-monitor for adjusting your monitor settings."));
+                        mNotification->setBody(tr("Use LXQt Monitor Settings to adjust your monitor settings."));
                         mNotification->update();
                         mNotification->setTimeout(1000);
                         mNotification->setUrgencyHint(LXQt::Notification::UrgencyLow);
--- a/lxqt-config-monitor/loadsettings.cpp
+++ b/lxqt-config-monitor/loadsettings.cpp
@@ -94,7 +94,8 @@ void LoadSettings::applyBestSettings()
                             KScreen::SetConfigOperation(config).exec();
                         qDebug() << "lxqt-config-monitor: Extended mode has been applied";
 
-                        mNotification->setSummary(tr("Default monitor settings have been applied. Use lxqt-config-monitor for adjusting your monitor settings."));
+                        mNotification->setSummary(tr("Default monitor settings applied."));
+                        mNotification->setBody(tr("Use lxqt-config-monitor for adjusting your monitor settings."));
                         mNotification->update();
                         mNotification->setTimeout(1000);
                         mNotification->setUrgencyHint(LXQt::Notification::UrgencyLow);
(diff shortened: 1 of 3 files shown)


stefonarch:
immagine


Somehow the body text is now too redundant which twice "Monitor Settings". The title can'e be changed?

tsujan:
> Dropping the second "monitor"?

 I don't thing "monitor" is redundant; anything less might be confusing.

> The title can't be changed?

It's set to the app's name in `liblxqt`. Adding an option for it requires making changes to `liblxqt`.

tsujan:
I got what you meant. IMO, it's too short and a little confusing, and it doesn't fix any real problem either.

stefonarch:
It's not about shortening, it's about saying "say Hello to say hello". That a monitor setting tool is for setting the monitor is obvious. so no second "monitor" is needed,

 We could also be more explicit:

"Invalid or missing configuration for this setup. Use LXQt Monitor Settings to adjust your settings."

tsujan:
> it's about saying "say Hello to say hello"

It isn't so (→ my previous comment).

Anyway, why are we discussing it here? I just split the original message in addition to saving size because I didn't have time to make another PR — bad practice. Isn't it better to do this somewhere else?

stefonarch:
If you'd agree to drop the redundant "monitor" ;)

More related to the PR: the "Saved Settings" window doesn't remember it's size, could that be done too?

Schermata_20240910_203216

tsujan:
> If you'd agree to drop the redundant "monitor" ;)

If I drop my monitor, I won't see anything; I have only one ;)

I'll have no problem with changing the text in another PR and after discussing it in a report.

> More related to the PR: the "Saved Settings" window doesn't remember its size, could that be done too?

It should be easy. But to track changes or probable regressions, let's limit this PR to its job.

tsujan:
A completely irrelevant question:

Does the app work under kwin_wayland as well as under X11? I don't have access to my external monitor yet.

stefonarch:
I've to say I tested _only_ in kwin_wayland... it looks perfectly working, inclusing saved settings (no message if for both situation a configuration is present). But we lack the "scale" for each monitor feature which is a deal breaker on wayland  now, there is an issue open about.

stefonarch (approved):

isf63:
This isn't working for me with Xfwm4 on X11. I haven't been able to test Wayland as I only have labwc, which does not support these settings.

On Xfwm4 it flashes the correctly remembered size for a few milliseconds and then resizes to the minimum size (unable to manually resize smaller).

tsujan:
That simple line doesn't depend on Wayland or X11. Something may be wrong with Xfwm4 or, maybe, its settings.

tsujan:
I'm checking it....

stefonarch:
Confirmed for openbox and kwin on x11.

tsujan:
I can also reproduce it under X11. My guess is that there's a problem that, for some mysterious reason, isn't triggered under Wayland: the dialog can't be resized on X11! I still don't know why...

tsujan:
OK, found it: https://github.com/lxqt/lxqt-config/blob/0eab8b6e3c97d12857c8ac27e12a5f5db9a74e33/lxqt-config-monitor/monitorsettingsdialog.cpp#L153

I may also have found why it seemed OK on Wayland: the `finished` signal of KScreen is triggered instantly on Wayland.

Will make a PR.

tsujan:
Removed that line in https://github.com/lxqt/lxqt-config/pull/1046.

On second though, the difference between Wayland and X11 doesn't seem to be related to KScreen but to Qt.

PR #1044, opened 2024-09-10
 Cleanup to Wayland message

isf63:
Same as https://github.com/lxqt/lxqt-globalkeys/pull/300.

### DIFF
--- a/lxqt-config-input/lxqt-config-input.cpp
+++ b/lxqt-config-input/lxqt-config-input.cpp
@@ -35,7 +35,7 @@
 int main(int argc, char** argv) {
     LXQt::SingleApplication app(argc, argv);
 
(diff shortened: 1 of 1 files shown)


isf63:
The change to the main message may not be worth needing re-translation.

stefonarch:
> The change to the main message may not be worth needing re-translation.

It will remove existing translations, I prevented it the other time but don't remember, will check.

stefonarch:
Done here https://github.com/lxqt/lxqt-config/pull/1045 thanks!
I preferred to keep "the", or do we use "your compositor" in other places too?

isf63:
"the" is fine. It's not worth needing new translations.

You forgot "Unsupported" being capitalized.

stefonarch:
Oops - edited your comment instead of quoting.
Ddidn't scroll down your PR, done now.

PR #1046, opened 2024-09-12
Fixed the size of monitor dialog

tsujan:
As a side note, every change should be checked under both X11 and Wayland, *even if it seems elementary*. In this case `QWidget::adjustSize()` didn't have effect under Wayland, and that made me think that the dialog should be resized correctly under X11 too, while it wasn't.


stefonarch (approved):
Now it's fine everywhere, openbox/kwin x11 and wayland.
Note to self: test x11 session always too.

PR #1047, opened 2024-09-13
lxqt-config-monitor: Change dialog button from QToolButton to QPushButton

isf63:
Change the "Settings" dialog button to a QPushButton. This simplifies the Designer UI file and fixes text/icon alignment (for example on Fusion.)

### DIFF
--- a/lxqt-config-monitor/monitorsettingsdialog.cpp
+++ b/lxqt-config-monitor/monitorsettingsdialog.cpp
@@ -86,7 +86,7 @@ MonitorSettingsDialog::MonitorSettingsDialog() :
 
     });
 
-    connect(ui.settingsButton, &QToolButton::clicked, this, &MonitorSettingsDialog::showSettingsDialog);
+    connect(ui.settingsButton, &QPushButton::clicked, this, &MonitorSettingsDialog::showSettingsDialog);
 
     // also make sure that widgets are shown completely (a Qt bug under Wayland?)
     LXQt::Settings settings(QStringLiteral("lxqt-config-monitor"));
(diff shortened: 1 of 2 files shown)


isf63:
liblxqt uses QDialogButtonBox which itself uses QPushButton.

Re:

>and fixes text/icon alignment (for example on Fusion.)

QToolButton has different (and in my opinion, worse) styling than QPushButton.

buttons

tsujan:
> fixes text/icon alignment

I'm afraid there's nothing to fix. It's like saying, "Don't use tabs because they look ugly with Fusion."

isf63:
QToolButton is styled differently in _most_ Qt styles. Logically dialog buttons should be consistently QPushButton as that is what liblxqt uses.

tsujan:
IMO, in no way should a code avoid using a widget because of how it's drawn by this or that Qt style.

isf63:
Apart from that, it simplifies the Designer UI and makes consistent use of QPushButton in dialog boxes. No functional changes other than its drawing by Qt styles.

I think these benefits are pretty uncontroversial.

isf63:
The correct QWidget is QPushButton here, QToolButton is mainly for toolbar buttons or for non-icon buttons.

Also please see Kvantum, the style is inconsistent there too.

tsujan:
> Also please see Kvantum, the style is inconsistent there too.

?!

isf63:
QToolButton has a visual style distinct from QPushButton - it's subtle but noticeable. I was saying this is also visible in Kvantum themes.

tsujan:
Oh, I thought you meant something's wrong with Kvantum ;)

There's nothing wrong with how a tool-button is drawn. And there's no inconsistency, because this particular button is different from the other ones. Being a tool-button, it's distinguished from the others. Also, if you make it a push-button, clicking it will make it the default button, and I'm sure that the original coder didn't want that.

PR #1050, opened 2024-09-17
lxqt-config-appearance: In palette dialog, disable OK button when != 1 palettes are selected

### LINKED ISSUE #1048: lxqt-config-appearance: Possible to select multiple palettes and then apply
Under 'Qt Palette' it is possible to click "Load Palette", select multiple palettes, and then hit OK. When loading with multiple selected, I believe only the last one is applied, but in general it's nondeterministic.

##### Expected Behavior
Allowing for selecting more than one is necessary, to enable deleting multiple at a time.

Not sure what to do with "OK".

##### Current Behavior
Selecting multiple palettes and hitting OK either applies the last one or has unknown behavior.

isf63:
Closes #1048.

### DIFF
--- a/lxqt-config-appearance/styleconfig.cpp
+++ b/lxqt-config-appearance/styleconfig.cpp
@@ -30,6 +30,7 @@
 #include "ui_palettes.h"
 #include <QDebug>
 #include <QStyleFactory>
+#include <QPushButton>
 #include <QToolBar>
 #include <QSettings>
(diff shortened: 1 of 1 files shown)


tsujan:
A user has no reason to select multiple items to load a palette, and even if he does that to test the app, it'll pass his test (→ https://github.com/lxqt/lxqt-config/issues/1048#issuecomment-2354273092).

Issue #1049, opened 2023-01-22
lxqt-config-appearance will start/replace xsettingsd even if GTK theme management is disabled


M4he:
The LXQt appearance configuration dialog (`lxqt-config-appearance`) will start an `xsettingsd` instance even if its GTK theme management option is disabled.
This breaks any custom `xsettingsd` configurations a user might be running, because it will terminate those (only one `xsettingsd` instance can exist at a time).

##### Expected Behavior

If "Set GTK themes" is disabled in `lxqt-config-appearance`, opening up the appearance configuration dialog should *not* start a new `xsettingsd` instance or replace an existing one that wasn't started by `lxqt-config-appearance`.

##### Current Behavior

Even if "Set GTK themes" is disabled in `lxqt-config-appearance`, simply opening the appearance configuration dialog will start a new `xsettingsd` instance as:

```
/usr/bin/xsettingsd -c /tmp/lxqt-config-appearance.<random-string>
```
  
If an `xsettingsd` instance was already running (e.g. configured by the user) before opening the dialog, it is replaced by the dialog's.
When the appearance dialog is closed, its `xsettingsd` instance is also terminated¹. In case the user had a custom `xsettingsd` instance running before, it is not restored, effectively leaving the system running without an active instance¹.

¹ ~~note: is this actually a bug of its own? I don't see the purpose of a dedicated `xsettingsd` instance if it terminates after closing the dialog. It has to keep running to be effective.~~ (EDIT: see my comments below)

##### Possible Solution

a) check if the option "Set GTK themes" is set at the startup of `lxqt-config-appearance` and don't start an `xsettingsd` instance unless it's explecitly enabled

or

b) check if an `xsettingsd` instance is already running (which wasn't started by `lxqt-config-appearance`) and if so, do not replace it.

##### Steps to Reproduce (for bugs)

1. start a custom `xsettingsd` instance
    - e.g. `xsettingsd -c ~/.xsettingsd` (see [here](https://github.com/derat/xsettingsd/wiki/Installation#configuration) about configuration)
2. open up `lxqt-config-appearance`
3. observe your `xsettingsd` instance getting replaced (check your process manager)
4. close the appearance configuration dialog
5. observe that no `xsettingsd` instance remains running

##### Context

Users may want to use a custom `xsettingsd` instance (e.g. started via LXQt's autostart) containing custom tweaks and don't want LXQt to interfere.

[...]

PR #1051, opened 2024-09-17
Start `xsettingsd` on X11 only when GTK themes should be set

tsujan:
Closes https://github.com/lxqt/lxqt-config/issues/1049

### DIFF
--- a/lxqt-config-appearance/configothertoolkits.cpp
+++ b/lxqt-config-appearance/configothertoolkits.cpp
@@ -100,7 +100,18 @@ ConfigOtherToolKits::ConfigOtherToolKits(LXQt::Settings *settings,  LXQt::Settin
 {
     mSettings = settings;
     mConfigAppearanceSettings = configAppearanceSettings;
+}
+
+ConfigOtherToolKits::~ConfigOtherToolKits()
+{
+    mXsettingsdProc.close();
+}
+
+void ConfigOtherToolKits::startXsettingsd()
+{
     if(QGuiApplication::platformName() != QStringLiteral("xcb"))
+        return;
+    if(QProcess::NotRunning != mXsettingsdProc.state())
         return;
     if(tempFile.open()) {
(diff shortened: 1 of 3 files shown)


M4he:
I have not tested this code (currently have no means to) and just looked at the changed code but I wondered:

If the dialog is opened while the ControlGTKThemeEnabled option is disabled, then xsettingsd is not started (check does not trigger in `showAdvancedOptions`). This is desired.

But if the user now clicks the checkbox for "Set GTK themes" for the first time and then "Apply", the GTK files will be written but is xsettingsd started as a result? To me it seems it is not (only when ControlGTKThemeEnabled was true before the dialog opened).

I'd expect that if the checkbox is enabled and "Apply" is clicked in the same dialog instance, xsettingsd gets started (if not already running) to apply the changes to running applications immediately.

tsujan:
> But if the user now clicks the checkbox for "Set GTK themes" for the first time

`xsettingsd` is started as soon as the box is checked. That can happen when the user checks it or when the dialog is shown with `ControlGTKThemeEnabled=true` in `~/.config/lxqt/lxqt-config-appearance.conf` (the code checks it).

M4he:
> > But if the user now clicks the checkbox for "Set GTK themes" for the first time
> 
> `xsettingsd` is started as soon as the box is checked. That can happen when the user checks it or when the dialog is shown with `ControlGTKThemeEnabled=true` in `~/.config/lxqt/lxqt-config-appearance.conf` (the code checks it).

Okay sorry then I might have misinterpreted the code. I only saw the starting happen in `showAdvancedOptions` via `initControls` which I thought would only be called once when the dialog is initially opened.

tsujan:
→ `GTKConfig::showAdvancedOptions` in the patch.

stefonarch (approved):
Works as expected.

tsujan:
@stefonarch, thanks for the test! I think the code is simple enough to be merged.

Issue #996, opened 2024-05-16
Add scale configuration under when underwayland


stefonarch:
##### Expected Behavior

Also scale should be configurable under kwin_wayland
##### Current Behavior

It's not.

##### Context

Different scale settings for different monitors can be useful in LXQt Kwin_wayland

stefonarch:
If monitors are scaled (using `systemsettings`) `lxqt-config-monitor` can't handle it:
b.com/lxqt/lxqt-config/issues/996, and atm only KDE can configure scaling.

![Schermata_20240921_102446](https://github.com/user-attachments/assets/eabba6d0-ce67-4bfb-a835-d7e8f2b6dc9a)
![Schermata_20240921_102511](https://github.com/user-attachments/assets/8211bc91-47df-4d81-8917-acb465db7079)

In this case at next login it will apply its configuration which is

![Schermata_20240921_210818](https://github.com/user-attachments/assets/c845098d-bf5b-47d0-b828-bdff9d1628b3)

and the monitor setup is completely broken (cursor can't exit primary monitor) until it's fixed by KDE's `systemsettings` (where it sees a space between monitors which is invalid.

tsujan:
Theoretically, we just need to add `KScreen::Output::scale` and `KScreen::Output::setScale` to the GUI.

stefonarch:
And the previews of the monitors will be fine too in their size?

tsujan:
Making previews work with scaling is only one part of the job. We need to add GUI elements to set and get the scales of all monitors, depending on whether we're on Wayland or not. We also need to save the new settings in a backward compatible way.

And these steps are just what came to my mind without writing a single line of code — sometimes there's a long way from theory to practice ;) But it's good to know that there is a way.

stefonarch:
Besides scaling also "This is my primary monitor" should be changed under kwin_wayland.

PR #1054, opened 2024-09-20
Scale monitor images on resizing monitor settings dialog

tsujan:
Also, the images are made a little larger by default.

Closes https://github.com/lxqt/lxqt-config/issues/1053

### DIFF
--- a/lxqt-config-monitor/monitorpicture.cpp
+++ b/lxqt-config-monitor/monitorpicture.cpp
@@ -26,6 +26,8 @@
 #include <QRectF>
 #include <KScreen/Mode>
 #include <QScrollBar>
+#include <QResizeEvent>
+#include <QTransform>
 
 #include "configure.h"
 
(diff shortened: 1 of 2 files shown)


stefonarch (approved):
Works fine. But we have an issue with scaled monitors, as we don't consider it. If `lxqt-config-monitor` is used with scaled monitors we will get a space in `systemsettings`  which complains about "space is not supported". But this belongs to https://github.com/lxqt/lxqt-config/issues/996, and atm only KDE can configure scaling.
Schermata_20240921_102446
Schermata_20240921_102511

tsujan:
> But we have an issue with scaled monitors

That's a completely different topic. I think the proper place for the above screenshots is https://github.com/lxqt/lxqt-config/issues/996. There they could show what we should deal with.

tsujan:
BTW, I'm going to add a commit for cleaning up the SVG image `monitor.svg`; will explain in its comment why such a cleanup is good.

stefonarch:
I just wanted to check if the monitor images itself work on scaled monitors (they do).

tsujan:
> ... (they do).

In a bad way.

PR #1057, opened 2024-09-30
Capitalize title "Platform unsupported"

isf63:
`<source>Platform Unsupported</source>` already exists in lxqt-config. This obsoletes `<source>Platform unsupported</source>`.


stefonarch:
Translations are separated. It will need preserving existing translations after.

PR #1076, opened 2024-11-03
styleconfig.ui: mark icon sizes as nontranslatable

Fat-Zer:


tsujan:
16 → ۱۶, 32 → ۳۲, etc.

Fat-Zer:
> 16 → ۱۶, 32 → ۳۲, etc.

@tsujan, Then a comment might be in order because as for now everybody just translates it as "16", "32" (including arabic).

Besides that, if you adamant to use localized numbers, it probably would be more productive to just fill the combo-box with `QLocale::toNumber()` instead.

tsujan:
Recently we've learned that Qt doesn't support all languages that can be translated and appear in the GUI.

>  (including arabic).

Those aren't Arabic ;)

Fat-Zer:
ok... won't insist...

> Those aren't Arabic ;)

sure, those were Urdu and Farsi respectively... and in my comment "1" was French and "6" was German =)

tsujan:
Those numbers aren't used in the code; their indexes are. Qt Designer doesn't have an option for localizing them, and adding localized numbers in the code just to prevent translations seemed like an overdoing.

But yes, removing those numbers from the ui file and adding localized ones in the code makes sense too (Qt is responsible for unsupported languages).

Now it's too late; the next release will be made the day after tomorrow.

PR #1081, opened 2024-12-10
Fixed the "Set Position" page in lxqt-config-monitor

tsujan:
The latest version of Qt 6.8 showed a problem in the code: the "Set Position" page was visible behind the "Fast Menu" page on startup, although everything worked fine.

The reason was that `MonitorPictureDialog` was a `QDialog` that was added to a stacked widget. The problem was solved by changing it to a `QWidget`. However, I didn't change its name to `MonitorPictureWidget`, because I didn't want to make a messy patch.

Also, a memory leak is fixed inside the same class.

P.S.
The main issue was found by @stefonarch.

### DIFF
--- a/lxqt-config-monitor/monitorpicture.cpp
+++ b/lxqt-config-monitor/monitorpicture.cpp
@@ -65,7 +65,7 @@ void MonitorPictureProxy::updatePosition()
 }
 
 MonitorPictureDialog::MonitorPictureDialog(KScreen::ConfigPtr config, QWidget * parent, Qt::WindowFlags f) :
-    QDialog(parent,f)
+    QWidget(parent,f)
 {
     updatingOk = false;
     firstShownOk = false;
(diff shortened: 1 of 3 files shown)


isf63:
I would test too, but `lxqt-config-monitor` periodically fails to launch on Arch lxqt -git. I recompiled all LXQt dependencies of lxqt-config but I think it has to do with a libkscreen update.

stefonarch (approved):
Tabs are fine now. I see I can't use _unified view_ for some reason.

tsujan:
@isf63 
You don't need git libkscreen for git LXQt. Here, `lxqt-config-monitor` starts without problem under LXQt+kwin_wayland and LXQt+X11 (of course, I always have git LXQt).

>  I can't use unified view for some reason.

I think I tried it last night and it worked  (but I was sleepy). But it needs a separate investigation.

tsujan:
BTW, if `lxqt-config-monitor` really has a problem with git libskreen, we might encounter it in the next version of libskreen, and then our libskreen codes will need an update.

isf63:
For libkscreen I'm using the main Arch repos, not -git. Now monitor settings are launching, maybe when libkscreen updates lxqt-config-git needs a recompile AND a session restart.

tsujan:
> maybe when libkscreen updates lxqt-config-git needs a recompile AND a session restart.

No recompilation is required (the only LXQt component that needs recompilation when Qt's version changes is libqtxdg), but a session restart may be needed.

stefonarch:
Probably unifying outputs isn't possible with mine, but Hyprland can do it, with black bars on top and bottom of the smaller one.

tsujan:
Yes, it's done in our code, inside https://github.com/lxqt/lxqt-config/blob/d91eadc17d8c4d88a0fe3d8bf87175a01867979e/lxqt-config-monitor/fastmenu.cpp#L56
Its code looks suspicious to me, although it should work fine.

PR #1086, opened 2025-01-02
Fixed clicking rectangles in Configuration Center

tsujan:
Previously, depending on the window size and the arrangement of items, the lower parts of some clicking rectangles might be removed, resulting in no action when those lower parts were clicked.

The reason was that the logic of what is called "binary search" in the code works fine only if the grid rectangles are considered, not the item rectangles.

NOTE: At first I wanted to replace "binary search" with an ordinary search because it worked only with a grid. However, since we do set a grid in the code, I've kept its beautiful logic.

### DIFF
--- a/src/qcategorizedview/qcategorizedview.cpp
+++ b/src/qcategorizedview/qcategorizedview.cpp
@@ -168,6 +168,10 @@ QPair<QModelIndex, QModelIndex> QCategorizedView::Private::intersectingIndexesWi
         itemRect.topLeft().rx() += horizontalOff;
         itemRect.bottomRight().ry() += verticalOff;
         itemRect.bottomRight().rx() += horizontalOff;
+        if (hasGrid()) { // see QCategorizedView::visualRect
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
If you can reproduce the clicking problem — which happened here with Kvantum Manager's item when there were only 3 columns and with `icon-size: 48px` — please test this; if not, never mind, because it may not be easy to reproduce for all.

stefonarch:
I can't reproduce it nowhere.

tsujan:
> I can't reproduce it nowhere (openbox,hyrpland) in no way.

It's OK. I was lucky that I saw one form of it. If it was always reproducible, we'd receive many reports. But the bug always existed.

tsujan:
According to tests, this fixes randomly cut texts that @fadi123go and @isf63 encountered, as well as randomly cut clicking rectangles that I encountered, both cases being about cutting from below. Also, its logic is simple. Merging…

PR #1087, opened 2025-01-05
Use `iso-flags-svg` for flags in Locale Config

tsujan:
KDE removed all flag images from KF6 and started to use `KCountryFlagEmojiIconEngine`, belonging to `KGuiAddons`, which doesn't seem acceptable as a dependency for LXQt. Instead, this simple patch uses `iso-flags-svg` and searches inside `$XDG_DATA_DIRS/iso-flags-svg/country-4x3/` for SVG flag images, if existing.

NOTE: IMHO, making it configurable would be an overkill, because flags aren't necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ belonging to LXQt, too.
 
 Binary `lxqt-config-locale`.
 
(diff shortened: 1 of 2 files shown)


stefonarch (approved):
GTM. Funny that nobody noticed it, flags went missing when we migrated to Qt6 no?

tsujan:
I think flags were missing after some old KDE5 package was removed. If so, some users might have never seen them.

PR #1089, opened 2025-01-09
Removed three cases of redundant writing to config files

tsujan:
… when LXQt Appearance Configuration was launched, and when it or Configuration Center was closed without a size change.


tsujan:
Straightforward. Merging…

Issue #1091, opened 2025-01-24
Add short time format to LXQt Locale Config


tsujan:
Although using a GUI tool alone for setting locales is not recommended in general (→ https://github.com/lxqt/lxqt/wiki/SettingLocale#note), LXQt Locale Config can serve for checking formats, but the short time format is missing from it. Any objection to adding it? Like:
![Image](https://github.com/user-attachments/assets/a37770da-6fa2-4f80-a5db-3c93c3226a80)

PR #1092, opened 2025-01-24
Added time short format to examples of LXQt Locale Config

tsujan:
Closes https://github.com/lxqt/lxqt-config/issues/1091

### DIFF
--- a/lxqt-config-locale/localeconfig.cpp
+++ b/lxqt-config-locale/localeconfig.cpp
@@ -407,7 +407,8 @@ void LocaleConfig::updateExample()
     }
 
     const QString numberExample = nloc.toString(1000.01);
-    const QString timeExample = tloc.toString(QDateTime::currentDateTime());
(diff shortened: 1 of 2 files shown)


stefonarch (commented):
immagine

The alignment isn't good IMO

isf63:
What about two rows, "Time" and "Short time", which is more clear but more verbose. Maybe there's more conventional phrasing for those for locales.

tsujan:
There's no need to 2 rows — the simpler, the better — and "Time" is perfectly aligned, showing that both lines belong to it.

Chiitoo:
It does look a bit odd to me too, even though yes, perfect in a way, I think I might more expect it to be like so:

```
Ora: venerdì [...]
     24/01/25 [...]
```

tsujan:
OK, tastes differ, and you won because you're the majority ;)

stefonarch:
Every time @tsujan changes his mind a new star is born ;)

tsujan:
I'm not obstinate; only logical ;) But this wasn't related to logic.

stefonarch (approved):
immagine

Issue #1093, opened 2025-01-25
Design flaw of LXQt Locale Config


tsujan:
LXQt Locale Config writes to two files when saving locales: `lxqt-config-locale.conf`, and `session.conf`. The first file is only for remembering the settings the next time the app is launched, while the second file will export the needed EVs and set the locales in the next LXQt sessions.

If the user selects the "No change" item in a combo-box, the app will remove its key from `lxqt-config-locale.conf`. So far, so good.

However, the way the app writes to `session.conf` is problematic:

 * The writing is done only when the main combo-box is anything other than "No change", in which case,
 * If "Detailed Settings" is checked, the EVs of the other combo-boxes that are different from "No change" will also be written, while the other EVs (if any) won't be touched; otherwise,
 * All EVs will be taken from the main combo-box and written to `session.conf`.

This has three problems:

 1. There is no way to remove all EVs from `session.conf` by using this GUI. In other words, there is no way to restore the system locale, other than removing those EVs from LXQt Session Settings → Environment. IMO, an app that adds EVs should also be able to remove them.
 2. There is no way to know the system locales from the GUI once the app is used for setting locales.
 3. The app doesn't tell the user that he should restart the session for the new locales to take effect.

1 and 2 are more important because they can trap the average user in a situation, where he doesn't know how to use the system locale for his LXQt session *after using LXQt Locale Config*.

tsujan:
The app does inform the user about the next session in the end. I removed that from the list of problems. But the main issue exists: No way for knowing the current locales or **not** setting a locale.

PR #1094, opened 2025-01-25
Improved locale setting

tsujan:
The following changes are made:

 * The first item of each list is "Unset", which means that LXQt won't set the corresponding locale with the next session. It serves as a cleaner.
 * The second item shows the locale of the *current* session ("Current: …"), which may be different from saved locales. It may not be the system default either (because session locales can be set in various ways).
 * A separator is added after the current locale. Its look depends on the Qt widget style.
 * When the app is launched, the combos show their saved or current values, depending on whether they are saved or not.

Closes https://github.com/lxqt/lxqt-config/issues/1093

### DIFF
--- a/lxqt-config-locale/localeconfig.cpp
+++ b/lxqt-config-locale/localeconfig.cpp
@@ -56,7 +56,7 @@ const static QString lcLanguage = QStringLiteral("LANGUAGE");
 LocaleConfig::LocaleConfig(LXQt::Settings* settings, LXQt::Settings* session_settings, QWidget* parent) :
     QWidget(parent),
     m_ui(new Ui::LocaleConfig),
-    hasChanged(new bool),
+    hasChanged(false),
     mSettings(settings),
     sSettings(session_settings)
 
@@ -69,8 +69,6 @@ LocaleConfig::LocaleConfig(LXQt::Settings* settings, LXQt::Settings* session_set
              << m_ui->comboCurrency
              << m_ui->comboMeasurement
              << m_ui->comboCollate;
-
-    hasChanged = false;
 
     initControls();
 }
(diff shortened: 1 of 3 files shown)


stefonarch:
There's an issue with the size, text about time (Central Europe...) is truncated now and at the first attempt of resizing it jumps to the correct size, both on labw and Niri (floating window):


https://github.com/user-attachments/assets/5293949e-a4be-4344-8fc4-b24ab201e25a

tsujan:
Strange! I didn't touch it. Are sure you sure this is new?

tsujan:
Here, the window is resized when I change "Time:" and the text becomes longer. My guess is that you hadn't tested it before because you had no reason ;)

Window resizing may be different on different Wayland compositors. Labwc is OK, but I clearly remember that sometimes windows weren't resized correctly on Wafire. I think it's a compositor issue.

stefonarch:
Open the app under wayland → text is cut off, try to resize a bit and the window jumps at the right size, at least under kwin, labwc and Niri. Under x11 it's fine, before it was fine too. My guess: it's exact the amount of "current". 

I don't think "current" is much needed as an info as  every setting shows "current" values always.

tsujan:
First, what you see about the size isn't related to this PR.

Second, "Current:" is one of the two reasons behind this PR. See https://github.com/lxqt/lxqt-config/issues/1093

> every setting shows "current" values always.

No, it showed (and still shows) the values you'd saved. If you saved and didn't restart the session, the GUI didn't let you know what the current locales were. "Current:" comes as the second item and before the separator for this reason.

tsujan:
> My guess: it's exact the amount of "current".

No. "Current:" is inside the line-edit, not on the example label. As I said, the example labels aren't touched by this PR.

> Under x11 it's fine

Which shows it's a Wayland issue.

tsujan:
BTW, with my font, I was able to see the lack of resizing under Labwc by using `Арассыыйа - саха тыла (sah_RU)`. But the window *is* resized when I change the time combo and put it on that locale. So, it's about resizing *on showing*. We had similar cases in libfm-qt, for which I added a workaround. A workaround may be possible for this too, but not in this PR.

stefonarch:
> First, what you see about the size isn't related to this PR.

It is caused by this PR, hope in your one-liner.

I see a similar issue with unsaved items when they are long (no "current"), the first letters are truncated (that is when reopening). Will add a second screenshot after session restart.

immagine

tsujan:
> It is caused by this PR

There's absolutely no way that this PR could **cause** it; at most, it has triggered it. And the single-line workaround will be in a file that isn't touched by the PR.

tsujan:
> the first letters are truncated…

If you mean inside combo-boxes, you're talking about something as remote from this PR as possible.

The shown text of an editable combo-box can be as long as needed, and so, it may not fit into the available space. What's important is the *popup*.

I equipped the app with a searching capability several years ago for this reason.

stefonarch (approved):
Looks fine, waiting for the next PR to fix the startup display.

tsujan:
Here you focused on the appearance rather than functionality. The next PR will be about appearance and resource usage.

PR #1095, opened 2025-01-27
Optimized LXQt Locale Configuration

tsujan:
1. Previously, each flag icon was created at least 6 times, for six combo-boxes. That caused a considerable delay on launching. Now each icon is created only once and is cached.
 2. The size policy of combo-boxes is changed to what Qt recommends for large lists, namely, `QComboBox::AdjustToMinimumContentsLengthWithIcon`.
 3. A truncated time label is prevented in the Examples box on showing the dialog under Wayland.
 4. The prompt dialog is now the child of the main dialog, such that a Wayland compositor like Labwc centers the former in front of the latter (other compositors may need to learn from Labwc).

Because of 2, there may be a delay the first time the list of a combo-box is shown by the user, but (1) there will be no delay afterward, and (2) the user even doesn't need to show the list — searching by typing was implemented years ago. Previously, that delay happened on launching.

### DIFF
--- a/lxqt-config-locale/localeconfig.cpp
+++ b/lxqt-config-locale/localeconfig.cpp
@@ -156,7 +156,9 @@ QString LocaleConfig::getCurrentforCombo(QComboBox *combo)
 void LocaleConfig::initCombo(QComboBox *combo, const QList<QLocale> & allLocales)
 {
     combo->clear();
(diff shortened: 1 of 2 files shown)


stefonarch (approved):
That was needed. Faster loading and truncated texts fixed,

PR #1096, opened 2025-01-27
lxqt-config-appearance: Add colon to remaining labels

isf63:
There were only three labels missing colons in lxqt-config-appearance. As far as the rest of lxqt-config goes, it looked like there weren't any more.

**Steps:**
1. Modify (three) .ui strings
2. `sed -i 's/<source>Qt Style<\/source>/<source>Qt Style:<\/source>/g' translations/*`
3. `sed -i 's/<source>GTK 2 Theme<\/source>/<source>GTK 2 Theme:<\/source>/g' translations/*`
4. `sed -i 's/<source>GTK 3 Theme<\/source>/<source>GTK 3 Theme:<\/source>/g' translations/*`
5. Manualy append colons in the exact manner as https://github.com/lxqt/libfm-qt/pull/1033
   (Except for zh_* which used a full-width colon.)

### DIFF
--- a/lxqt-config-appearance/gtkconfig.ui
+++ b/lxqt-config-appearance/gtkconfig.ui
@@ -74,7 +74,7 @@
             <item row="1" column="0">
              <widget class="QLabel" name="label_3">
               <property name="text">
-               <string>GTK 2 Theme</string>
+               <string>GTK 2 Theme:</string>
               </property>
              </widget>
             </item>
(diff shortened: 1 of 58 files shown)


tsujan:
It makes sense — I mean adding those three colons; don't know about translations, which @stefonarch might want to check.

stefonarch:
It looks GTM

isf63:
I checked it again with grep.

PR #1097, opened 2025-01-27
Appearance - GTK Style: Move information label out of QFormLayout

isf63:
Move information label outside of the QFormLayout.

It looked weird being aligned to one column.


tsujan (approved):
Good idea.

PR #1103, opened 2025-02-08
Replace `qMin`,`qMax`,`qBound`,`qRound`,`qAbs` with standard funcs

isf63:
This is the last one of these for a while

### DIFF
--- a/liblxqt-config-cursor/xcr/xcrthemefx.cpp
+++ b/liblxqt-config-cursor/xcr/xcrthemefx.cpp
@@ -88,7 +88,7 @@ QList<XCursorThemeFX::tAnimSeq> XCursorThemeFX::parseScript (const QString &scri
         qWarning() << "script error (frame):" << s;
         continue;
       }
-      a.from = std::clamp(a.from, (quint32)1, maxFrame) - 1;
+      a.from = std::max(std::min(maxFrame, a.from), (quint32) 1) - 1;
       a.to = a.from;
     } else {
       // a..b
@@ -97,13 +97,13 @@ QList<XCursorThemeFX::tAnimSeq> XCursorThemeFX::parseScript (const QString &scri
         qWarning() << "script error (frame from):" << s;
         continue;
       }
-      a.from = std::clamp(a.from, (quint32)1, maxFrame) - 1;
+      a.from = std::max(std::min(maxFrame, a.from), (quint32) 1) - 1;
       if (!str2num(fld[0].mid(hyph+1), a.to)) {
        qDebug() << "script error (frame to):" << s;
         qWarning() << "script error (frame to):" << s;
         continue;
       }
-      a.to = std::clamp(a.to, (quint32)1, maxFrame) - 1;
+      a.to = std::max(std::min(maxFrame, a.to), (quint32) 1) - 1;
     }
     // delay
     if (!str2num(fld[1], a.delay)) {
--- a/liblxqt-config-cursor/xcr/xcrthemefx.cpp
+++ b/liblxqt-config-cursor/xcr/xcrthemefx.cpp
@@ -19,6 +19,8 @@
 #include <QStringList>
 #include <QTextStream>
 
+#include <algorithm>
+
 #include "xcrimg.h"
 #include "xcrxcur.h"
 #include "xcrtheme.h"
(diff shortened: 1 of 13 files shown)


liblxqt-config-cursor/xcr/xcrthemefx.cpp
```diff
@@ -86,7 +88,7 @@ QList<XCursorThemeFX::tAnimSeq> XCursorThemeFX::parseScript (const QString &scri
         qWarning() << "script error (frame):" << s;
         continue;
       }
-      a.from = qMax(qMin(maxFrame, a.from), (quint32)1)-1;
+      a.from = std::clamp(a.from, (quint32)1, maxFrame) - 1;
```
tsujan, line 91: The behavior of `clamp` is undefined if `maxFrame` is less than `1`. So, it's better to revert to `max` and `min`. Besides, `algorithm` is already included.

liblxqt-config-cursor/xcr/xcrthemefx.cpp
```diff
@@ -95,13 +97,13 @@ QList<XCursorThemeFX::tAnimSeq> XCursorThemeFX::parseScript (const QString &scri
         qWarning() << "script error (frame from):" << s;
         continue;
       }
-      a.from = qMax(qMin(maxFrame, a.from), (quint32)1)-1;
+      a.from = std::clamp(a.from, (quint32)1, maxFrame) - 1;
```
tsujan, line 100: Ditto.

liblxqt-config-cursor/xcr/xcrthemefx.cpp
```diff
@@ -95,13 +97,13 @@ QList<XCursorThemeFX::tAnimSeq> XCursorThemeFX::parseScript (const QString &scri
...
       if (!str2num(fld[0].mid(hyph+1), a.to)) {
        qDebug() << "script error (frame to):" << s;
         qWarning() << "script error (frame to):" << s;
         continue;
       }
-      a.to = qMax(qMin(maxFrame, a.to), (quint32)1)-1;
+      a.to = std::clamp(a.to, (quint32)1, maxFrame) - 1;
```
tsujan, line 106: Ditto.

liblxqt-config-cursor/xcr/xcrthemexp.cpp
```diff
@@ -145,10 +147,10 @@ static bool readNextSection (QTextStream &stream, CursorInfo &info) {
     bool numOk = XCursorThemeFX::str2num(nv[1].trimmed(), num);
     if (!numOk) num = 0;
     if (name == QLatin1String("frames")) {
-      info.frameCnt = qMax(num, (quint32)1);
+      info.frameCnt = std::max(num, (quint32)1);
       info.wasFrameCnt = true;
     } else if (name == QLatin1String("interval")) {
-      info.delay = qMax(qMin(num, (quint32)0x7fffffffL), (quint32)10);
+      info.delay = std::clamp(num, (quint32)10, (quint32)0x7fffffffL);
```
tsujan, line 153: Needs including `cmath` — or is it included elsewhere?

tsujan: Oh, sorry, it doesn't need cmath.

liblxqt-config-cursor/xcr/xcrthemefx.cpp
isf63: I'll look into it... Nested `max`, `min` constructs were confusing to me, but I  understand how they might differ from clamp.

tsujan: > I'll look into it

Thanks.

> Nested max, min constructs were confusing to me

Sometimes they're needed. I encountered a similar case in featherpad. We didn't have to be cautious with `qBound`.

BTW, I saw more KDE sources, in which your changes were made. It seems that they're preparing their codes for the removal of `qMin`, `qMax`, etc. from Qt. However, I don't think that will happen soon.

isf63: `std::*` is the much cleaner solution, despite `std::clamp` having shortcomings. I think Qt is doing the right thing - it may compile faster and having other optimizations in compilers.

tsujan (approved):
GTM

liblxqt-config-cursor/xcr/xcrthemefx.cpp
tsujan: At some point, we'll have to get rid of `qreal` too, but it's too soon now.

Issue #924, opened 2023-02-27
Keyboard repeat setting shows always 0 0  launch


stefonarch:
When started:

![screen-27 feb_19:28](https://user-images.githubusercontent.com/10681413/221652513-2cc46b38-9d8c-4b2a-9a88-86769b11e4ac.jpg)

As soon a value is sligthly changed it shows the real thing:

![screen-27 feb_19:29](https://user-images.githubusercontent.com/10681413/221652506-abb55d73-2f47-47fd-bf00-98b8aea52603.jpg)

Save, close and reopen and back at screenshot 1.

##### Expected Behavior

Show always current value

##### Steps to Reproduce (for bugs)

As above.

##### Context

Wondering why I saw a  `Long 0` at the end of the line.  Adding  `ms` should  be done too.

##### System Information

* Distribution & Version: debian 12
* Package version:  1.2.0

tsujan:
Closed in https://github.com/lxqt/lxqt-config/commit/94577098e7683fc29c5b37a17e6e9ed1944e5f3c

PR #1107, opened 2025-02-22
Let lxqt-config-input be shown partially under Wayland

tsujan:
… because some of its settings are applicable to Qt apps.

### DIFF
--- a/lxqt-config-input/keyboardconfig.cpp
+++ b/lxqt-config-input/keyboardconfig.cpp
@@ -61,6 +61,14 @@ KeyboardConfig::KeyboardConfig(LXQt::Settings* _settings, QSettings* _qtSettings
   connect(ui.keyboardBeep, &QAbstractButton::clicked, this, &KeyboardConfig::settingsChanged);
   connect(ui.cursorFlashTime, QOverload<int>::of(&QSpinBox::valueChanged), this, &KeyboardConfig::settingsChanged);
   connect(ui.keyboardNumLock, &QAbstractButton::clicked, this, &KeyboardConfig::settingsChanged);
+
+  if (QGuiApplication::platformName() == QLatin1String("wayland"))
+  { // disable the settings that don't work under Wayland
+    ui.keyboardDelay->setEnabled(false);
+    ui.keyboardInterval->setEnabled(false);
+    ui.keyboardBeep->setEnabled(false);
+    ui.keyboardNumLock->setEnabled(false);
+  }
 }
 
 KeyboardConfig::~KeyboardConfig() {
@@ -70,10 +78,12 @@ KeyboardConfig::~KeyboardConfig() {
 void KeyboardConfig::initControls() {
   ui.keyboardDelay->blockSignals(true);
   ui.keyboardDelay->setValue(delay);
+  ui.label_delay->setNum(delay);
   ui.keyboardDelay->blockSignals(false);
 
   ui.keyboardInterval->blockSignals(true);
   ui.keyboardInterval->setValue(interval);
+  ui.label_interval->setNum(interval);
   ui.keyboardInterval->blockSignals(false);
 
   ui.keyboardBeep->setChecked(beep);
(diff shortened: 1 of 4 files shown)


stefonarch (commented):
1. I can't see double click interval doing anything.
2. Cannot change character repeat, it's blocked.
3. Maybe we need to show still an info box before opening it.
4. I still think we shouldn't show cursor in 2 different places.

tsujan:
> I can't see double click interval doing anything.

You should, in Qt apps.

> Cannot change character repeat, it's blocked.

Any setting related to X11 is disabled.

> Maybe we need to show still an info box before opening it.

Don't make it complex.

> I still think we shouldn't show cursor look in 2 different places.

Not related to this PR.

tsujan:
> I can't see double click interval doing anything.

If you meant to ask how you can see its effect, type a few characters in a line inside any Qt app (e.g. FeatherPad) and double click to select all of them. Change the number to 1400 and see how slow you could do it. The default is 400 milliseconds.

tsujan:
> Cannot change character repeat, it's blocked.

Actually, this is a very good point. Although this setting is for X11 in `lxqt-config`, Qt supports `KeyboardAutoRepeatRate` since 6.5, but we haven't enabled it in `lxqt-qt-plugin`.

So, I'll make a PR to enable it in `lxqt-qt-plugin`, and then, I'll also enable it here by adding a commit.

stefonarch:
Ok, I thought it was mandatory to need that time for the double click, never used that setting as I use single click.

For the keyboard delays: it should be greyed or hidden if it doesn't work, now it looks like a bug.

And I think still we shouldn't expose a half working (Qt only) cursor theme setting if we have a fully working one.

tsujan:
> For the keyboard delays: it should be greyed or hidden if it doesn't work,

Gray=disabled. The X11 sliders are already disabled. There's no reason to disable their texts too.

> And I think still we shouldn't expose a half working (Qt only) cursor theme setting

This needs a separate discussion — not here.

For auto-repeating, I'll make a PR and request a review from you after adding a commit to this one.

tsujan:
The problem with auto-repeating is that it's already saved in `session.conf` for lxqt-session, but now we need it in `lxqt.conf` for Qt. I should find a way to solve this inconsistency in a backward compatible way…

stefonarch:
> Gray=disabled. The X11 sliders are already disabled. There's no reason to disable their texts too.

It's not visible here that anything is disabled.

>For auto-repeating, I'll make a PR and request a review from you after adding a commit to this one.

For disabling?

tsujan:
> It's not visible here that anything is disabled in autorepeat.

If you mean the entry that says "Type in the following box…", it's not a setting but just a widget for testing. If you mean something else, please attach a screenshot.

As for auto-repeat in Qt, I checked Qt's source. It's used only in spin-boxes when the up/down key is kept pressed, not in any text entry. In other words, Qt's implementation is incomplete, at best. I even patched `lxqt-qt-plugin` to test: there was no change in any Qt text entry, and the behavior of spinboxes was inconsistent, although it *partially* reflected the value I set for auto-repeat.

Therefore, this PR doesn't need a change, and there's no reason to make a PR for `lxqt-qt-plugin`.

However. tell me what you mean by autorepeat not disabled, preferably with a screenshot.

stefonarch:
immagine
 I didn't check that autorepeat is disabled here.

tsujan:
Oh, that's a problem in Fusion :)

OK, I'll make the texts as well as that entry disabled too; Fusion couldn't show its problem with them.

tsujan:
Done. Please test again — with the same Fusion palette ;)

stefonarch (approved):
1. I can't see double click interval doing anything.
2. Cannot change character repeat, it's blocked.
3. Maybe we need to show still an info box before opening it.
4. I still think we shouldn't show cursor in 2 differen

stefonarch:
Closes https://github.com/lxqt/lxqt-config/issues/924

And adding ms after the numbers?

tsujan:
> And adding ms after the numbers?

Good idea. Will do it here…

> Some explanation for all the greyed out things would be fine to add.

I thought about it but, since there were two places, adding explanatory notes would clutter the code as well as the GUI. Let's not do that for something that might change later. Disabled widgets are enough to know what can't be done under Wayland.

tsujan:
Adding an `ms` there needs a code change because the code uses [`QLabel::setNum`](https://doc.qt.io/qt-6/qlabel.html#setNum). Maybe in another PR.

PR #1108, opened 2025-02-23
Apply cursor changes to GTK in `lxqt-config-input` if needed

tsujan:
"if needed" means if the GTK option is enabled in LXQt Appearance Configuration.

The patch also shows info on mouse/touchpad/keyboard settings under Wayland and, if the GTK option isn't enabled, a hint on GTK.


isf63:
After a session restart it changes the cursor for GTK3 and Qt programs. I was unable to test GTK2 as I don't have any programs installed that require it.

stefonarch (approved):
Works fine.

PR #1113, opened 2025-03-05
Fixed screen images with different scales in monitors page

tsujan:
This fix wasn't needed on X11, because all screens had the same scale factor there, but its lack can cause a gap between screens on Wayland (only kwin_wayland for now).

See https://github.com/lxqt/lxqt-config/issues/996 for the other things that need to be fixed.

### DIFF
--- a/lxqt-config-monitor/monitorpicture.cpp
+++ b/lxqt-config-monitor/monitorpicture.cpp
@@ -194,7 +194,8 @@ MonitorPicture::MonitorPicture(QGraphicsItem * parent,
 {
     this->monitorWidget = monitorWidget;
     this->monitorPictureDialog = monitorPictureDialog;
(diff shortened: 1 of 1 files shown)


stefonarch (approved):
Fixes the issue on loading settings if one or both screen is/are scaled. After changing scaling in kwin the settings needs to be saved again.

tsujan:
> After changing scaling in kwin the settings needs to be saved again.

Yes, because it doesn't have a setting for scaling yet. It's a shame that even if we add it, it'll be only for kwin_wayland. Oh, X11 should be taken care of too; I don't think different scale factors will work there.

tsujan:
It shouldn't be about writing to compositors' config files but handling Wayland outputs directly, as an app like kanshi does.

PR #1116, opened 2025-03-18
Cleaned up the code of fast options in LXQt Monitor Settings

tsujan:
This part of the code seemed like a draft that wasn't cleaned up later.

### DIFF
--- a/lxqt-config-monitor/fastmenu.cpp
+++ b/lxqt-config-monitor/fastmenu.cpp
@@ -41,7 +41,7 @@ FastMenu::FastMenu(KScreen::ConfigPtr config, QWidget* parent) :
 
     ui.setupUi(this);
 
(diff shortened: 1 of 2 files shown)


tsujan:
@stefonarch 
Particularly, the unified view should select the largest common size of all monitors.

To be sure, you could buy 5-6 monitors with different resolutions and test the fast menu with all of them connected.

stefonarch:
You think I'll get refunded by our OpenCollective sponsor for that?
immagine

tsujan:
Ah, that would be an ideal setup for testing everything ;)

stefonarch (commented):
All looks working fine here, but actually both my monitors have the same size so I can't tell.

tsujan:
Thanks for testing it. As for monitor sizes, the code is essentially the same as the previous one; I just removed a redundant loop and cleaned it up.

tsujan:
OK, we can't fully test this, but since its code is simple enough, I take full responsibility and merge it. After all, no computer will explode because of it.

PR #1118, opened 2025-03-29
lxqt-config-appearance: Add GTK 4 theming

IdfbAn:
Two things I should say:
1. X11 has been tested working by me, but Wayland (and thus GSettings?) hasn't been tested.
2. Opening `gtk4-demo` gives me this warning:
```
Error setting gtk-cursor-theme-size in /home/REDACTED/.config/gtk-4.0/settings.ini:
Key file contains key “gtk-cursor-theme-size” in group “Settings” which has a value that cannot be interpreted.
```
The theme _seems_ to still work regardless, but if this actually causes any problems, and/or anyone knows how to fix this, it'd be great.

### DIFF
--- a/lxqt-config-appearance/gtkconfig.cpp
+++ b/lxqt-config-appearance/gtkconfig.cpp
@@ -41,6 +41,7 @@ GTKConfig::GTKConfig(LXQt::Settings *configAppearanceSettings, ConfigOtherToolKi
     connect(ui->advancedOptionsGroupBox, &QGroupBox::clicked, this, &GTKConfig::settingsChanged);
     connect(ui->gtk2ComboBox, QOverload<int>::of(&QComboBox::activated), this, &GTKConfig::settingsChanged);
     connect(ui->gtk3ComboBox, QOverload<int>::of(&QComboBox::activated), this, &GTKConfig::settingsChanged);
+    connect(ui->gtk4ComboBox, QOverload<int>::of(&QComboBox::activated), this, &GTKConfig::settingsChanged);
 }
 
 GTKConfig::~GTKConfig()
@@ -54,6 +55,7 @@ void GTKConfig::initControls()
     // Fill themes
     QStringList gtk2Themes = mConfigOtherToolKits->getGTKThemes(QStringLiteral("2.0"));
     QStringList gtk3Themes = mConfigOtherToolKits->getGTKThemes(QStringLiteral("3.*"));
+    QStringList gtk4Themes = mConfigOtherToolKits->getGTKThemes(QStringLiteral("4.*"));
 
     if(!mConfigAppearanceSettings->contains(QStringLiteral("ControlGTKThemeEnabled")))
         mConfigAppearanceSettings->setValue(QStringLiteral("ControlGTKThemeEnabled"), false);
@@ -66,9 +68,11 @@ void GTKConfig::initControls()
     // Fill GTK themes
     ui->gtk2ComboBox->addItems(gtk2Themes);
     ui->gtk3ComboBox->addItems(gtk3Themes);
+    ui->gtk4ComboBox->addItems(gtk4Themes);
 
     ui->gtk2ComboBox->setCurrentText(mConfigOtherToolKits->getGTKThemeFromRCFile(QStringLiteral("2.0")));
     ui->gtk3ComboBox->setCurrentText(mConfigOtherToolKits->getGTKThemeFromRCFile(QStringLiteral("3.0")));
+    ui->gtk4ComboBox->setCurrentText(mConfigOtherToolKits->getGTKThemeFromRCFile(QStringLiteral("4.0")));
 
     update();
 }
@@ -77,8 +81,13 @@ void GTKConfig::applyGTKStyle()
 {
     if (ui->advancedOptionsGroupBox->isChecked())
     {
+        // GTK4
+        QString themeName = ui->gtk4ComboBox->currentText();
+        mConfigOtherToolKits->setGTKConfig(QStringLiteral("4.0"), themeName);
+        if(QGuiApplication::platformName() == QStringLiteral("wayland"))
+            mConfigOtherToolKits->setGsettingsConfig(QStringLiteral("4.0"), themeName);
         // GTK3
-        QString themeName = ui->gtk3ComboBox->currentText();
+        themeName = ui->gtk3ComboBox->currentText();
         mConfigOtherToolKits->setGTKConfig(QStringLiteral("3.0"), themeName);
         if(QGuiApplication::platformName() == QStringLiteral("wayland"))
             mConfigOtherToolKits->setGsettingsConfig(QStringLiteral("3.0"), themeName);
--- a/lxqt-config-appearance/gtkconfig.ui
+++ b/lxqt-config-appearance/gtkconfig.ui
@@ -83,12 +83,6 @@
             </item>
             <item row="1" column="0">
              <widget class="QLabel" name="label_4">
-              <property name="sizePolicy">
-               <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
-                <horstretch>0</horstretch>
-                <verstretch>0</verstretch>
-               </sizepolicy>
-              </property>
               <property name="text">
                <string>GTK 3 Theme:</string>
               </property>
@@ -96,6 +90,22 @@
             </item>
             <item row="1" column="1">
              <widget class="QComboBox" name="gtk3ComboBox"/>
+            </item>
+            <item row="2" column="0">
+             <widget class="QLabel" name="label_5">
+              <property name="sizePolicy">
+               <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+                <horstretch>0</horstretch>
+                <verstretch>0</verstretch>
+               </sizepolicy>
+              </property>
+              <property name="text">
+               <string>GTK 4 Theme:</string>
+              </property>
+             </widget>
+            </item>
+            <item row="2" column="1">
+             <widget class="QComboBox" name="gtk4ComboBox"/>
             </item>
            </layout>
           </item>
--- a/lxqt-config-appearance/configothertoolkits.cpp
+++ b/lxqt-config-appearance/configothertoolkits.cpp
@@ -67,6 +67,17 @@ gtk-cursor-theme-name = %6
 gtk-cursor-theme-size = %7
 )GTK3_CONFIG";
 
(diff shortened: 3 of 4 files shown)


tsujan:
The code is written neatly. Thanks!

> Opening gtk4-demo gives me this warning:

This needs to be investigated. Perhaps  `gtk-cursor-theme-size` doesn't  work with GTK4 and another string should be used?

I don't use any GTK4 app and have forgotten everything about GTK after more than a decade of not dealing with it. So, my code review is far from enough. Thorough tests are needed.

IdfbAn:
> > Opening gtk4-demo gives me this warning:
> 
> This needs to be investigated. Perhaps `gtk-cursor-theme-size` doesn't work with GTK4 and another string should be used?

It actually [does](https://docs.gtk.org/gtk4/property.Settings.gtk-cursor-theme-size.html) exist in GTK 4, which is why this is so weird.

stefonarch:
Installed some gtk4 apps, get the same warning but actually can't set any dark theme (adwaita-dark, breeze-dark), neither in X11 nor in Wayland.

IdfbAn:
> Installed some gtk4 apps, get the same warning but actually can't set any dark theme (adwaita-dark, breeze-dark), neither in X11 nor in Wayland.

Hmm, interesting. HighContrastInverse works for me, could you try that?

stefonarch:
> Hmm, interesting. HighContrastInverse works for me, could you try that?

Tried also that, same, it seems, should be dark no? `gnome-themes-extra` was installed already.

IdfbAn:
Hold on a second, are you sure the GTK apps you're using don't use libadwaita?

stefonarch:
> Hold on a second, are you sure the GTK apps you're using don't use libadwaita?

I'm not really familiar with GTK4, I tried just Qeegie, Nautilus and Amberol, could well be. An example?

IdfbAn:
> I'm not really familiar with GTK4, I tried just Qeegie, Nautilus and Amberol, could well be. An example?

Nautilus and Amberol use libadwaita, but Geeqie doesn't. I've just installed Geeqie to test it, but it seems like it actually uses GTK 3 (at least on Arch, btw).

tsujan:
Why testing with GTK4 apps when `gtk4-widget-factory` (belonging to `gtk4-demos`) exists?

For some reason, I have `~/.config/gtk-4.0/settings.ini` as
```
[Settings]
gtk-application-prefer-dark-theme=true
gtk-cursor-theme-name=ComixCursors-White
gtk-cursor-theme-size=24
gtk-decoration-layout=close,minimize,maximize:icon
gtk-enable-animations=true
gtk-font-name=Sans 9
gtk-icon-theme-name=Plataro-GTK
gtk-primary-button-warps-slider=false
```
And this screenshot shows `gtk4-widget-factory` in front of `gtk3-widget-factory` here, under Wayland (labwc):

gtk4

But, I remember that I used `dconf-editor` too, when GTK4 came out.

The warning message I see is different and isn't our concern:
```
Gdk-WARNING **: 11:20:21.388: vkAcquireNextImageKHR(): A swapchain no longer matches the surface properties exactly, but can still be used to present to the surface successfully. (VK_SUBOPTIMAL_KHR) (1000001003)
```
As far as I know, GNOME devs didn't like the idea of a themeable GTK. So, maybe they've removed themeablity from GTK4 — just a guess …

IdfbAn:
> Why testing with GTK4 apps when `gtk4-widget-factory` (belonging to `gtk4-demos`) exists?

I've already tested that too, and it also works for me. Can you explicitly set the theme to Adwaita-dark to see if it also works in Wayland?

tsujan:
No, setting `gtk-theme-name=Adwaita-dark` inside `~/.config/gtk-4.0/settings.ini` doesn't work, but `gtk-application-prefer-dark-theme=true` does.

However, setting `Adwaita-dark` inside `dconf-editor → /org/gnome/desktop/interface/gtk-theme → Custom value` works.

tsujan:
The irony is that `dconf-editor` depends on GTK3, not GTK4. GTK is a mess!

lxqt-config-appearance/gtkconfig.ui
```diff
@@ -109,6 +109,17 @@
             </item>
            </layout>
           </item>
+          <item>
+           <widget class="QLabel" name="gtk4ThemeDisclaimerLabel">
+            <property name="text">
+             <string>The selected GTK 4 theme does not apply to applications that use Libadwaita.
```
tsujan, line 115: IMHO, this is redundant because the user doesn't need to know what Adwaita is, let alone Libadwaita.

IdfbAn: This is more for the (potential) people that'll install an app that uses Libadwaita (even if they don't know), and get confused when their theme doesn't work on it.

tsujan: Even a GNOME tool like `dconf-editor` doesn't talk about "libadwaita". But if you think it's needed, please use a tooltip instead.

stefonarch:
I still can't get anything in dark...
Transmission-gtk is GTK4 without libadwaita, but ignores settings.

tsujan:
Oh, neither Adwaita nor Adwaita-dark doesn't support GTK4. At least here, they belong to `gnome-themes-extra` and their directories inside `/usr/share/themes/` don't include any `gtk-4.0` directory.

So, I think the tooltip ("Doesn't apply to applications that use Libadwaita") may be misleading. Switching to a theme that isn't for GTK4 shouldn't do anything.

IdfbAn:
> Switching to a theme that isn't for GTK4 shouldn't do anything.

Doesn't that mean these themes are being falsely included in the GTK 4 box?

tsujan:
> Doesn't that mean these themes are being falsely included in the GTK 4 box?

It seems so — if they're provided by `gnome-themes-extra` in Arch — because a theme should support GTK4 in order to be applicable to GTK4 apps.

tsujan:
> Adwaita-dark does work for me …

I don't have the illusion that GNOME's behavior is logical, but if it is, you may have an overriding theme in a user folder. If you don't have any theme with a `gtk-4.0` folder, then blame GNOME devs for the inconsistent behavior. At least, what @stefonarch and I see exonerates GNOME ;)

IdfbAn:
Alright then, so what do we do? Do I just remove the tooltip, or do I need to do a bit more? Ignoring the cursor theme, of course. (Which I probably shouldn't lol)

tsujan:
> Do I just remove the tooltip

IMHO, the tooltip doesn't convey correct info. So, I think, yes, it can be removed.

> do I need to do a bit more?

Something should be done because it doesn't work for @stefonarch, and so, it might not work for many users. But I don't know what or how. As I mentioned earlier, I'm not a GTK expert; what I said so far was only logical. For starters, if a theme isn't applicable to a version of GTK, it shouldn't be shown for it. Now, whether GNOME should do it (as Qt does) or we should, I don't know …

IdfbAn:
If there's a problem with my latest commit, let me know.

IdfbAn:
> It should be tested again, especially with a theme that supports GTK4 (is there any at all?).

I've just tested Orchis-Dark, and it does seem to work with the GTK demo apps. It gives me an "Expected a valid color" warning, but this may or may not be because of the theme itself.

tsujan:
> … it does seem to work with the GTK demo apps

Good.

> … but this may or may not be because of the theme itself.

Yes, we don't need to worry about GTK warnings; there are lots of them.

stefonarch:
Still nothing on my side also with orchid*. What worked was https://github.com/vinceliuice/WhiteSur-gtk-theme#fix-for-libadwaita-not-perfect

and it does:
```
$ cat .config/gtk-4.0/
assets/         gtk.css         gtk-dark.css    gtk-Dark.css    gtk-Light.css   windows-assets/ 

```
Only setting which looks working is font-size.

tsujan:
> Still nothing on my side also with orchid*.

Does the `orchid*` you tested have a `gtk-4.0` subfolder?

Also, Adwaita shouldn't be listed for GTK4 now. Is that the case?

tsujan:
I confirm that the keys `gtk-cursor-theme-name` and `gtk-cursor-theme-size` don't work *under Wayland, with GTK3 or GTK4, and regardless of this PR*. Maybe they should be tested only under X11.

stefonarch:
immagine
Adwaita isn't listed. Orchis-dark has a GTK4 subfolder.

tsujan:
> Cursor is read from one file

Yes, that's for the global cursor setting. I just wanted to say that we might need to consider the differences between Wayland and X11 when testing; @IdfbAn uses X11, you and I Wayland.

tsujan:
OK, I reread the PR and found the cause of the problem. It's simple: `gsettings` doesn't distinguish between GTK3 and GTK4. In `GTKConfig::applyGTKStyle()`, the PR first first sets it for GTK4 and then overrides it for GTK3.

IdfbAn:
> In `GTKConfig::applyGTKStyle()`, the PR first sets the theme for GTK4 and then overrides it for GTK3.

What's the exact line that's apparently problematic?

stefonarch:
Searching for info I found that using an env var `export GTK_THEME=Orchis-Dark` is working for all GTK4 apps here...

lxqt-config-appearance/gtkconfig.cpp
```diff
@@ -77,8 +81,13 @@ void GTKConfig::applyGTKStyle()
 {
     if (ui->advancedOptionsGroupBox->isChecked())
     {
+        // GTK4
+        QString themeName = ui->gtk4ComboBox->currentText();
+        mConfigOtherToolKits->setGTKConfig(QStringLiteral("4.0"), themeName);
+        if(QGuiApplication::platformName() == QStringLiteral("wayland"))
+            mConfigOtherToolKits->setGsettingsConfig(QStringLiteral("4.0"), themeName);
         // GTK3
```
tsujan, line 89: The call to `gsettings` (in `setGsettingsConfig`) will be repeated and overridden below, for GTK3.

I think GTK3 and GTK4 should have the same settings and be set by a single call to `gsettings`. If I'm right, a single combo-box should be used for both, the previous commit should be reverted, and some minor changes will be needed too.

What I say is based on what I see in `dconf-editor`: when I set it to `HighContrast`, both GTK3 and GTK4 themes change. The call to `gsettings` does the same.

This creates a confusing situation though: what if a theme doesn't support GTK4? I think, in that case, only the GTK3 theme will be applied.

ahsand97:
@IdfbAn are you still working on this?

PR #1122, opened 2025-04-23
"Properly" multiple values desktop entry keys

luis-pereira:
Terminating multiple values keys with a semicolon is optional. It's current practice though.

See https://github.com/lxqt/lxqt-session/pull/589


luis-pereira:
Superseded by https://github.com/lxqt/lxqt-config/pull/1124

PR #1125, opened 2025-05-17
Add an icon to 'Open theme folder'

isf63:
This was part of a bigger experiment to add "Open icons folder" and "Open cursors folder" context menu and double click, as https://github.com/lxqt/lxqt-config/pull/854 does for LXQt themes. The complexity seemed too much, but may be do-able in the future.

This PR trivially adds an icon.

### DIFF
--- a/lxqt-config-appearance/lxqtthemeconfig.cpp
+++ b/lxqt-config-appearance/lxqtthemeconfig.cpp
@@ -183,6 +183,7 @@ void LXQtThemeConfig::contextMenu(const QPoint& p)
 {
     QMenu menu;
     QAction *a = menu.addAction(tr("Open theme folder"));
+    a->setIcon(QIcon::fromTheme(QStringLiteral("document-open")));
     connect(a, &QAction::triggered, [this, p] {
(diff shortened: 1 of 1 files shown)


tsujan:
For reasons beyond this comment, the dialog of LXQt Appearances Configuration doesn't and can't follow symbolic icon colorization (although it enables/disables it correctly). So, it's preferable to avoid icons in it as far as possible.

In short, the reason is that it has to load all icon themes in its second page, and that ruins colorization inside the whole dialog.

tsujan:
To make it clearer it with an image:

color

isf63:
I guess I'm striking out with my recent PRs :-]

If you find time, please review the remaining https://github.com/lxqt/lxqt-themes/pull/141. I will close the others fairly soon.

isf63:
This context menu (and potentially, if added, the other pages' "Open ___ folder" menus) shouldn't have icons for several additional reasons:
- It's a single item menu, so there is no needed differentiation.
- An icon would be a bit odd/distracting.

PR #1127, opened 2025-06-01
lxqt-config-input: Change default cursor size 16->24

isf63:
https://github.com/lxqt/lxqt-session/pull/569 changed `session.conf` `cursor_size=` from `18` to `24`. Still however `lxqt-config-input` tried to set (in the absence of a value?) `16`.

### DIFF
--- a/liblxqt-config-cursor/selectwnd.ui
+++ b/liblxqt-config-cursor/selectwnd.ui
@@ -78,7 +78,7 @@
       <number>128</number>
      </property>
      <property name="value">
-      <number>16</number>
+      <number>24</number>
      </property>
     </widget>
    </item>


isf63:
`session.conf` was set to 24 from the weird value of 18. `lxqt-config-input` recommends 16 (not desirable), I'm guessing in the case of not being able to read an existing value.

tsujan:
Oh, you've included the link here :)

> Still however lxqt-config-input tried to set (in the absence of a value?) 16.

16 is the default value of Qt, and because of that, it's chosen as the default in `lxqt-qtplugin` too.

tsujan (approved):
GTM

isf63:
Done. Unfortunately, I forgot to create a branch and merged it directly on master :/ :/

Issue #1128, opened 2025-06-06
2.2.0 is FTBFS on Fedora Rawhide


sfalken:
Looks like there's been a couple changes in kscreen and Qt6 that need updating.

This is building against libkscreen-6.3.91-2 and Qt-6.9.0

This is the Kscreen error in lxqt-config-monitor
```
[   50s] In file included from /usr/include/KF6/KScreen/kscreen/config.h:11,
[   50s]                  from /usr/include/KF6/KScreen/KScreen/Config:1,
[   50s]                  from /home/abuild/rpmbuild/BUILD/lxqt-config-2.2.0-build/lxqt-config-2.2.0/redhat-linux-build/lxqt-config-monitor/lxqt-config-monitor_autogen/EWIEGA46WW/../../../../lxqt-config-monitor/fastmenu.h:28,
[   50s]                  from /home/abuild/rpmbuild/BUILD/lxqt-config-2.2.0-build/lxqt-config-2.2.0/redhat-linux-build/lxqt-config-monitor/lxqt-config-monitor_autogen/EWIEGA46WW/moc_fastmenu.cpp:9,
[   50s]                  from /home/abuild/rpmbuild/BUILD/lxqt-config-2.2.0-build/lxqt-config-2.2.0/redhat-linux-build/lxqt-config-monitor/lxqt-config-monitor_autogen/mocs_compilation.cpp:2:
[   50s] /usr/include/KF6/KScreen/kscreen/output.h:577:14: error: declaration of ‘operator<=’ as non-function
[   50s]   577 |         auto operator<=>(const BpcRange &) const = default;
[   50s]       |              ^~~~~~~~
[   50s] /usr/include/KF6/KScreen/kscreen/output.h:577:22: error: expected ‘;’ at end of member declaration
[   50s]   577 |         auto operator<=>(const BpcRange &) const = default;
[   50s]       |                      ^~
[   50s]       |                        ;
[   50s] /usr/include/KF6/KScreen/kscreen/output.h:577:24: error: expected unqualified-id before ‘>’ token
[   50s]   577 |         auto operator<=>(const BpcRange &) const = default;
[   50s]       |                        ^
```

And this is the Qt6 Deprecated function in liblxqt-config-cursor

```
[...]

PR #1129, opened 2025-06-06
Silenced all new compilation warnings

tsujan:
Closes https://github.com/lxqt/lxqt-config/issues/1128


isf63:
Successfully silenced all warnings on Qt `6.8.3+dfsg-0ubuntu2`

Chiitoo:
All gone with Qt 6.9.2 (that is, 6.9 `git master`), and GCC 15.1.0 as well.

tsujan:
In the case of lxqt-panel, unfortunately, the `QVariant` warnings seem to happen inside a Qt source file related to QtDbus — I was able to silence only those lxqt-panel warnings that were related to `statusnotifieriteminterface.h`.

All in all, I think the `QVariant` warnings should be fixed in Qt.

stefonarch (approved):
All clean here too.

PR #1131, opened 2025-06-18
Appearance → Icons Theme: Add open theme folder UI

isf63:
lxqt-config-appearance → 'Icons Theme', double-click the QTreeWidget row of an icon set to open its installed directory.

This copies the functionality of the LXQt Theme page.

**Notes**

~Also copy the right-click menu that 'LXQt Theme' has (?)~ Done

Add this for 'Cursor' (?)


isf63:
Added a context menu and also a fix for the one in LXQt Theme

isf63:
I agree mostly, it was intended on being for navigating to installation paths in case finding `/usr/share/icons/...` was too obscure. But there is little use for navigating there.

tsujan:
IMHO, there's no need to add codes to our sources for opening folders of cursor themes or icon themes, because they're independent of DEs.

PR #1132, opened 2025-06-19
lxqt-config-input → Cursor: Remove misleading info label

isf63:
Setting the cursor writes to `.gtkrc-2.0` and `.config/gtk-3.0/settings.ini` - these files override current GTK2/3 theme settings. So there's no possibility of a GTK theme overriding what is set here, making the label misleading.


stefonarch:
The label is about the fact that those settings only have effect  if the checkbox in the GTK section in `lxqt-config-appearance` is checked - only then it will write to the GTK files. 
It was introduced on purpose.

isf63:
Then I misunderstood the sentence. Confirmed by unchecking the box and looking at `.config/gtk-3.0/settings.ini`.

Some things:
- It could be made more clear with "To also apply the changes to GTK, check LXQt Appearance Configuration → GTK Style → Set GTK themes."
- It doesn't show up on the Appearance → Cursor embedding.

stefonarch:
In the appearance settings is the section about GTK, so there is no need for it. Some font settings are applied too,

PR #1146, opened 2025-10-01
Assign application chooser to the parent window

City-busz:
This ensures that the dialog is attached to the main window. Also mark it as modal.


City-busz:
Why would the user open more than one instance of this dialog? It can be more confusing to have multiple application choosers opened at the same time, even for the same file type. The user just selects an application and closes the dialog.

PR #1147, opened 2025-10-01
Assign advanced settings to the parent window

City-busz:
This ensures that the dialog is attached to the monitor settings window.


City-busz:
When the Advanced Settings is opened, it already blocks the access to the root window, so the user has to close it even if it's minimized, before able to interact with or even close the root window. So I think there is no reason to provide the ability to minimize it separately from the main window.

PR #1163, opened 2025-12-12
Restore margins to Default Applications page

isf63:
The rationale is that, after observing the GUI carefully, Associations → right column, has margins.

Default Applications had removed margins, therefore the groupboxes were touching the scrollbar (at least in Fusion.) I see no reason to remove the margins.


isf63:
Yes, I think there are perhaps several more blocks that can be removed throughout the `mimetypeviewer.ui`. Qt Designer (which I only use for previewing) can often add strange things initially.

Issue #1135, opened 2025-02-02
[Feature request] Support acceleration profile


baaannnaaannn:
**Is your feature request related to a problem? Please describe.**
 need just setting for mouse acceleration          
 I'm always frustrated when I can't set acceleration to flat from gui           

**Describe the solution you'd like**
 please add in mouse options that really important acceleration profile option  
how it should work, explain the difference from current behavior.    
Just let change mouse acceleration  

 A screenshot might help.                                              Don't have one,just like in KDE maybe

**Describe eventual alternatives you've considered**

##### Context

Configure very first thing that I do,pls do its very important and basic function

PR #1165, opened 2025-12-18
X11: Added draglock  and disable-while-typing settings

stefonarch:
@isf63 @Chiitoo if you could test/review this?
@tsujan please take a look at the .ui.file.

I didn't  get where to connect (disable)  "Lock drag" when "tap and drag" is unchecked.

If this is good I'd like to add at least:
* ~3-finger-drag~ Wayland only
* accel-profile flat|adaptive (https://github.com/lxqt/lxqt-config/issues/1135)
* disable with external mouse
* disable while typing


tsujan:
> For the .ui it looks like…

@stefonarch and I will fix it soon (through Matrix).

stefonarch:
> It compiles, I have no touchpad available so before "Tap to click" and "Tap to drag" were disabled - and now "Tap and drag" and "Lock drag" (IMO they need better root names.)

You mean now "tap to click" is enabled even if you don't have a touchpad?

tsujan (approved):
I didn't find any problem in its code. So, GTM if it's tested and works fine.

I suggest waiting a day or two before giving it to git users.

PR #1167, opened 2025-12-27
lxqt-config-appearance: Remove double margins for scrollarea pages

isf63:
Previously the pages with scrollareas (Widget Style, GTK Config) had bigger encompassing margins due to the extra layouts. This also worked against needing the scrollareas.


tsujan:
I don't have a strong opinion. Will check it later.

PR #1168, opened 2025-12-31
Remove unnecessary headers

isf63:
These are remnants of code from probably many years ago and the headers served no purpose other than slowing down compilation.

In some cases I also reorganized header order to be:
```
#include "local.h"
..
#include <LXQt/...>
..
#include <Qt headers>
```

Testing on my machine after re-running `cmake ..`, compilation speeds increased 6.8%:
```
Before:
time make -j6
real    0m44.706s
user    3m47.929s
sys     0m23.574s
====================
After:
real    0m41.635s
user    3m38.059s
sys     0m22.413s
```

### DIFF
--- a/liblxqt-config-cursor/cfgfile.cpp
+++ b/liblxqt-config-cursor/cfgfile.cpp
@@ -7,6 +7,7 @@
  * To Public License, Version 2, as published by Sam Hocevar. See
  * http://sam.zoy.org/wtfpl/COPYING for more details.
  */
+
 #include "cfgfile.h"
 
 #include <QDir>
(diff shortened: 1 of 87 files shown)


tsujan:
Reviewing this PR needs time, especially because it's highly preferable to include the needed headers explicitly.

> which has particularly slow compilation.

I don't think the included headers have a tangible effect on the compilation time.

isf63:
>preferable to include the needed headers explicitly

The vast majority of removals were `<QFoo>`. I simply searched for the "QFoo" class name in the translation unit and if not found removed the header. Rest assured everything still compiles.

>I don't think the included headers have a tangible effect on compile time

Best estimates are 10ms - 50ms for Qt headers depending on the machine. Here I get >3 seconds shaved off.

isf63:
Closing this because it's impossible to review given numerous Qt versions used and different Linux distributions / BSDs / ...

PR #1169, opened 2025-12-31
Fixed remembering checked state of draglock checkbox

stefonarch:


tsujan:
And happy new year to everyone!

tsujan (approved):
LGTM. Please test it thoroughly before merging.

tsujan:
`TouchpadConfig::initFeatureControl` changes the checked state of `tapToDragEnabledCheckBox`, and then the enabled state of `dragLockEnabledCheckBox` should be changed because of the connection to the signal `QCheckBox::toggled`. The last change you made works but it does a redundant thing too, while the code that I gave GTM should have worked fine without a redundancy. So, I'm curious to know about which test you talked.

stefonarch:
I know, most settings are enabled only if "touchpad" is chosen in the combo.
 But if we don't connect them later you could enable draglock without enabling drag which is weird. 
Testing was always fine, except after removing the wrong line, where initial state was wrong and it got disabled only after toggling "drag" twice.

tsujan:
> … which is weird.

No, the converse is wrong. A disabled widget shows that the device doesn't support its corresponding feature.

> … where initial state was wrong and …

A result of the above. When a part of the logic is contradicted, inconsistencies are expected. If you suppress them, something else will happen. The correct approach is respecting the logic.

stefonarch:
You didn't test it. The initial state _when touchpad device was chosen_ was wrong. 
The initial state for other devices was fine.

tsujan:
> which "lines" do you mean?

→ https://github.com/lxqt/lxqt-config/pull/1169#issuecomment-3704329098

> This PR moves one line to the right place

This PR should add `initFeatureControl(ui.dragLockEnabledCheckBox, device.dragLockEnabled());`, as it did, and also remove those lines.

> The enabled state is set: …

The enabled state should be set ONLY based on what the device supports. Please don't create complications by breaking this — I explained some of those complications in my previous comment.

stefonarch:
You can check a feature which is not enabled (drag lock). I'll stop here.

isf63:
I see no issue in the code other than just needing to remove the other
`ui.dragLockEnabledCheckBox->setEnabled(ui.tapToDragEnabledCheckBox->isChecked());`
as `initFeatureControl` already handles enabled state.

isf63:
IMO groupboxes with one item should be avoided. The side-by-side checkboxes with dynamic enabled/checked state as I described present hardly any confusion.

tsujan:
→ https://github.com/lxqt/lxqt-config/pull/1171

stefonarch:
Where 2 lines of very nice and logical organized patchwork do work fine the changes with https://github.com/lxqt/lxqt-config/pull/1171 are:

1. Shows checked  (disabled) touchpad settings now for any device.

Before:
screen-01-03-18-03-03

Now:
screen-01-03-18-02-04

2. The same issue I want to fix with this PR - "Drag lock" is enabled wrongly when opening the app:

screen-01-03-18-01-27

tsujan:
Oh, there's a note in Qt doc: "… and you can explicitly enable individual children in a group box that is unchecked. **This is however not recommended, as it could create a surprising experience for the end user.**"

lxqt-config-input/touchpadconfig.cpp
```diff
@@ -60,7 +60,6 @@ TouchpadConfig::TouchpadConfig(LXQt::Settings* _settings, QWidget* parent):
     connect(ui.edgeScrollingRadioButton, &QAbstractButton::clicked, this, &TouchpadConfig::settingsChanged);
     connect(ui.buttonScrollingRadioButton, &QAbstractButton::clicked, this, &TouchpadConfig::settingsChanged);
     connect(ui.tapToDragEnabledCheckBox, &QCheckBox::toggled,ui.dragLockEnabledCheckBox, &QCheckBox::setEnabled);
```
tsujan: Please add a space after `&QCheckBox::toggled,` and also an empty line before this connection, because it's different from the others.

tsujan (approved):
Very good.

PR #1180, opened 2026-01-29
update for kscreen 6.6

sfalken:
the kscreen API removed the isPrimary() method from KScreen::Output in 6.6, Port to new method Config::PrimaryOutput()


stefonarch:
I don't understand the practical impact. 
Tested shortly under X11 and I can set the monitor I want to "primary" and it's applied and reopening monitor settings it's still there. As I started to use dual monitor setups only with Wayland I ignore what "primary" would be like if set to the external on the right, never got familiar with the idea of "primary".

I doubt that distros will do anything here, so  a point release is probably fine, it will go anyway only to the rolling ones I guess.

PR #1197, opened 2026-04-20
Cursor config: Add page title

isf63:


isf63:
This adds a "page title" to `liblxqt-config-cursor` which appears in **Keyboard and Mouse Settings** and also **LXQt Appearance Configuration**. In those configs all other pages had a title.

tsujan:
> … which appears in Keyboard and Mouse Settings and …

It's redundant here. Take a closer look at the GUI to know why.

isf63:
The "title" is redundant, or the page being in 2 configs? This PR initially was to reword the whole leading paragraph (which I think needs work; however, new translations don't come easy) _and_ add a title but I eliminated the new tr.

tsujan:
I was referring to the extra title "Cursor": it's redundant `infoLabel` tells everything clearly; it doesn't need rewording either.

I see no reason to change anything here.

isf63:
Alright.

I'll avoid "micro-polish" furthermore as I have done that for a while and am runnning out beneficial observations there.

PR #1198, opened 2026-04-21
Tidy applicationchooser.ui

isf63:
- Tidy the layout and remove some redundant tags

Before: before

After: after

I saw another layout issue when there is no MIME icon available, but didn't address it here.

### DIFF
--- a/lxqt-config-file-associations/applicationchooser.ui
+++ b/lxqt-config-file-associations/applicationchooser.ui
@@ -35,7 +35,7 @@
       <enum>QFrame::Raised</enum>
      </property>
      <layout class="QGridLayout" name="gridLayout">
-      <item row="1" column="0">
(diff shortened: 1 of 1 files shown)


stefonarch:
GTM

but I see this here, no "file manager"


immagine

isf63:
(Associations → inode/directory → Default application) shows no "File managers" section for me.

(Default Applications → File Manager → Change...) should, which is what page I was on.

isf63:
On the Associations page,  "MIME type" and "Default application" could use all-word capitalization (the standard for English titles,) but I know translations come in slow.

stefonarch:
> On the Associations page, "MIME type" and "Default application" could use all-word capitalization (the standard for English titles,) but I know translations come in slow.

I think it's possible preserve translations.

PR #1206, opened 2026-05-28
cmake remove nested project

luis-pereira:
- **Remove translations CMake artifact**
- **Remove nested CMake project**


stefonarch:
GTM

PR #1210, opened 2026-06-24
Silence Uic warning

isf63:


tsujan:
`label_2` already exists below. It's enough to change the second instance of `label` to `label_1` and leave the other labels as they are — as the warning suggested:

```
The name 'label' (QLabel) is already in use, defaulting to 'label1'.
```

isf63:
I seem to have suffered from some mysterious brain damage...

tsujan:
Don't underestimate these sweltering heat waves that are everywhere  ;)

PR #1214, opened 2026-07-23
Delete abandoned/unused `savesettings.cpp`

isf63:
This was the final old `Q_FOREACH`/`foreach` in lxqt/lxqt. A few instances still exist but compilation is skipped.


luis-pereira:
`savesettings.cpp` isn't built. `savesettings.h` doesn't exist.
What's it's purpose ?

isf63:
Oh, good question.

isf63:
[Here](https://github.com/lxqt/lxqt-config/commits/master/lxqt-config-monitor/savesettings.cpp) is the history for the file. 11 years ago: [savesettings.ui](https://github.com/lxqt/lxqt-config/blob/f994f156d9fc6c5e503e78a23fcbca4911da3de9/lxqt-config-monitor/savesettings.ui)

It looks like it was an earlier version of Monitor Settings → Advanced Settings (`managesavedsettings.{cpp,h,ui}`). Should be safe to remove.

tsujan:
> Should be safe to remove.

Yes. Let's get rid of it.

PR #1215, opened 2026-07-25
Brightness: Add Qt icons fallback to head label

isf63:
The icon besides the header text did not display with Breeze or Breeze Dark icons, only with GTK ones → Adwaita, GNOME.

I'm certainly open to finding more appropriate icons - "-symbolic" should generally be avoided.

### DIFF
--- a/lxqt-config-brightness/brightnesssettings.cpp
+++ b/lxqt-config-brightness/brightnesssettings.cpp
@@ -34,7 +34,8 @@ BrightnessSettings::BrightnessSettings(QWidget *parent):QDialog(parent)
     mMonitorsInitial = mBrightness->getMonitorsInfo();
     mBacklight = new LXQt::Backlight(this);
 
-    ui->headIconLabel->setPixmap(QIcon::fromTheme(QStringLiteral("display-brightness-symbolic")).pixmap(32, 32));
+    ui->headIconLabel->setPixmap(QIcon::fromTheme(QStringLiteral("video-display-brightness"),
+                                 QIcon::fromTheme(QStringLiteral("video-display"))).pixmap(32, 32));
 
     ui->backlightSlider->setEnabled(mBacklight->isBacklightAvailable() || mBacklight->isBacklightOff());
     ui->backlightGroupBox->setEnabled(mBacklight->isBacklightAvailable() || mBacklight->isBacklightOff());


tsujan:
>  "-symbolic" should generally be avoided.

I agree completely.

isf63:
How about
1. `video-display-brightness`
2. `video-display`

?

tsujan:
Yes, `video-display-brightness` can be the icon, and `video-display` the fallback.

tsujan (approved):
Good, IMO.

lxqt-config-brightness/brightnesssettings.cpp
```diff
@@ -34,7 +34,8 @@ BrightnessSettings::BrightnessSettings(QWidget *parent):QDialog(parent)
     mMonitorsInitial = mBrightness->getMonitorsInfo();
     mBacklight = new LXQt::Backlight(this);
 
-    ui->headIconLabel->setPixmap(QIcon::fromTheme(QStringLiteral("display-brightness-symbolic")).pixmap(32, 32));
+    ui->headIconLabel->setPixmap(QIcon::fromTheme(QStringLiteral("video-display-brightness"),
+                                 QIcon::fromTheme(QStringLiteral("video-display"))).pixmap(32, 32));
```
palinek, line 38: Hi, late to the party. But IMO it is worth mentioning.
The `video-display` used as a fallback to `video-display-brightness` is pointless. As the [xdg standard](https://specifications.freedesktop.org/icon-naming/latest/) already defines, that searching the icon should use also the "ascendant" icons, where the levels are separated by dash. So if `X-Y-Z` can't be found, then the search continues with `X-Y`, ... and with `X`

tsujan: Yes, I knew that, but the tool can also be used without LXQt's icon handling.

palinek: It is not LXQt specific icon handling. It is XDG standard.

isf63: Qt follows the `X-Y-Z` → `X-Y` → `X` part of the XDG icon standard? I never knew about it.

palinek: https://github.com/qt/qtbase/blob/dev/src/gui/image/qiconloader.cpp#L565-L575

isf63: If you think it's worth another PR let me know. I think it's okay currently, those two strings can be easily searched and verified to be contained in GTK and Qt sets.

